PR #8, opened 2019-08-19
PW-1560 official way to return ajax json response

cyattilakiss:
Use ajaxDie instead of echo and exit
Use \Tools::jsonEncode instead of json_encode


helper/Data.php
```diff
@@ -289,7 +289,7 @@ public function buildControllerResponseJson($action, $details = []) {
                 break;
         }
 
-        return json_encode($response);
+        return \Tools::jsonEncode($response);
```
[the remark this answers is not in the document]
msilvagarcia: As discussed with the team, we're going to support PHP 5.2+ for now. So let's keep using this.

controllers/front/Payment.php
```diff
@@ -33,6 +33,21 @@ public function __construct()
         $this->helper_data->startSession();
     }
 
+    /**
+     * @param null $value
+     * @param null $controller
+     * @param null $method
+     * @throws PrestaShopException
+     */
+    protected function ajaxRender($value = null, $controller = null, $method = null)
```
KadoBOT: Doesn't the FrontController contains this logic already? Why this is being repeated?

cyattilakiss:
@msilvagarcia Agree using the json_encode insted of the static method.

rikterbeek:
@msilvagarcia can you solve the conflict ?

controllers/front/Payment.php
rikterbeek: @KadoBOT is deprecated by Prestashop that is why we need to add this logic.

msilvagarcia: This method is only present in PrestaShop 1.7. I added a wrapper around that so if we drop support to 1.6 it's easy to remove this code.

PR #15, opened 2019-08-28
PW-1555 Create Adyen notification table

msilvagarcia:


adyen.php
```diff
@@ -113,6 +147,11 @@ public function install()
     public function uninstall()
     {
         // TODO: delete adyen configurations (api-key)
+        $db = Db::getInstance();
+        /** @noinspection SqlWithoutWhere SqlResolve */
+//        $db->execute('DELETE FROM `' . _DB_PREFIX_ . 'adyen_notification`');
```
Aleffio:
```suggestion

```

msilvagarcia: I should have let the command there actually. Thanks for picking this up.

PR #20, opened 2019-09-26
PW-1657 Transform 3DS2 dialog in a modal

msilvagarcia:


controllers/front/ThreeDSProcess.php
```diff
@@ -81,6 +81,7 @@ public function postProcess()
                     'message' => "3D secure 2.0 failed, payload details are not found"
                 ]
             ));
+            return;
```
cyattilakiss: after the exit is added, this is not required

controllers/front/ThreeDSProcess.php
```diff
@@ -98,6 +99,7 @@ public function postProcess()
                     'message' => '3D secure 2.0 failed'
                 ]
             ));
+            return;
```
cyattilakiss: after the exit is added, this is not required

controllers/front/ThreeDSProcess.php
```diff
@@ -112,6 +114,7 @@ public function postProcess()
                     'token' => $result['authentication']['threeds2.challengeToken']
                 ]
             ));
+            return;
```
cyattilakiss: after the exit is added, this is not required

views/templates/front/payment.tpl
```diff
@@ -225,7 +227,7 @@
                             type: "POST",
                             url: paymentProcessUrl,
                             data: data,
-                            dataType: "json",
+                            dataType: "text json",
```
cyattilakiss: Is there any case when we don't return json back but text?

views/templates/front/payment.tpl
```diff
@@ -296,10 +298,7 @@
                             type: "POST",
                             url: threeDSProcessUrl,
                             data: data,
-                            dataType: "json",
-                            done: function(response) {
-                                return response;
-                            }
+                            dataType: "text json"
```
cyattilakiss: Is there any case when we don't return json back but text?

views/templates/front/payment.tpl
msilvagarcia: The server response type is `text/html`. According to the jQuery docs, adding `text json` tells the library that the response needs to be converted from plain text to JSON. Adding this solved an issue with the 3DS2 modal for me.

views/templates/front/payment.tpl
msilvagarcia: The server response type is `text/html`. According to the jQuery docs, adding `text json` tells the library that the response needs to be converted from plain text to JSON. Adding this solved an issue with the 3DS2 modal for me.

PR #21, opened 2019-09-26
PW-1556 execute notifications

cyattilakiss:
- Add new admin controller which the cronjob can use to process
notifications
 - Rename notificationProcessor to notificationReceiver because it
 handles receiving and saving notifications
 - Fix for ajaxRender: exit after rendering
 - Fix for ajaxRender: set this->ajax to true by default
 - Fix for try catches: if catch triggered, return something ot exit to
 prevent errors
 - Create Configuration class for global settings
 - Fix for link generation: burnt in links are not working since
 friendly url setting overwrites the default url structure and triggers
 error when trying to open an 'old' link
 - add installTab and createCronJobToken to upgrade 1.0.1
 - process authorisation notifications - adding a message if processed
 - expose admin controller for cron jobs to call

### DIFF
--- a/adyen.php
+++ b/adyen.php
@@ -29,21 +29,16 @@ if (!defined('_PS_VERSION_')) {
 
 class Adyen extends \PaymentModule
 {
-    const TEST = 'test';
-    const LIVE = 'live';
-    const CHECKOUT_COMPONENT_JS_TEST = 'https://checkoutshopper-test.adyen.com/checkoutshopper/sdk/3.0.0/adyen.js';
-    const CHECKOUT_COMPONENT_JS_LIVE = 'https://checkoutshopper-live.adyen.com/checkoutshopper/sdk/3.0.0/adyen.js';
-    const CHECKOUT_COMPONENT_CSS_TEST = 'https://checkoutshopper-test.adyen.com/checkoutshopper/sdk/3.0.0/adyen.css';
-    const CHECKOUT_COMPONENT_CSS_LIVE = 'https://checkoutshopper-live.adyen.com/checkoutshopper/sdk/3.0.0/adyen.css';
-    const VERSION = '1.0.1';
-    const MODULE_NAME = 'adyen-prestashop';
-
-
+    /**
+     * Adyen constructor.
+     *
+     * @throws \Adyen\AdyenException
+     */
     public function __construct()
     {
         $this->name = 'adyen';
         $this->tab = 'payments_gateways';
-        $this->version = self::VERSION;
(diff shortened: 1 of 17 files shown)


controllers/front/Payment.php
```diff
@@ -102,10 +111,10 @@ public function postProcess()
 
         if (!\Validate::isLoadedObject($customer)) {
             if (empty($_REQUEST['isAjax'])) {
-                \Tools::redirect('index.php?controller=order&step=1');
+                \Tools::redirect($this->context->link->getPageLink('order', $this->ssl) . '?step=1');
             } else {
                 $this->ajaxRender($this->helperData->buildControllerResponseJson('redirect',
-                    ['redirectUrl' => 'index.php?controller=order&step=1']));
+                    ['redirectUrl' => $this->context->link->getPageLink('order', $this->ssl) . '?step=1']));
```
msilvagarcia: Do we need to add the question mark here?

cyattilakiss: we can pass the request as a parameter and then the function will decide how to concatenate.

service/notification/NotificationProcessor.php
```diff
@@ -0,0 +1,186 @@
...
+     */
+    protected function addMessage($notification)
+    {
+        $successResult = (strcmp($notification['success'], 'true') == 0 ||
+            strcmp($notification['success'], '1') == 0) ? 'true' : 'false';
+        $success = (!empty($notification['reason'])) ? "$successResult <br />reason:" . $notification['reason'] . PHP_EOL : $successResult . PHP_EOL;
+
+        $type = 'Adyen HTTP Notification(s):';
+        $comment = sprintf(
+            '%s <br /> eventCode: %s <br /> pspReference: %s <br /> paymentMethod: %s <br />' .
```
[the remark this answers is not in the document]
cyattilakiss: WORKS!!! :)

adyen.php
```diff
@@ -294,13 +342,21 @@ public function displayForm()
...
+                    'type' => 'password',
                     'label' => $this->l('HMAC key for notifications'),
                     'name' => 'ADYEN_NOTI_HMAC',
                     'size' => 20,
-                    'required' => true,
+                    'required' => false,
                     'hint' => $this->l('Must correspond to the notification HMAC Key in the Adyen Backoffice under Settings => Notifications => Additional Settings => HMAC Key (HEX Encoded)')
                 ),
+                array(
+                    'type' => 'text',
```
Aleffio: should it be 'password' like the other encrypted fields?

cyattilakiss: was thinking about it but we expose it in the descripiton under it so not necessary

PR #22, opened 2019-10-02
Tidying up the main class

msilvagarcia:
This basically fixes most of the warnings from PhpStorm in the `adyen.php` file.

### DIFF
--- a/adyen.php
+++ b/adyen.php
@@ -21,18 +21,48 @@
  */
 
 // init composer autoload
+use Adyen\AdyenException;
+use Adyen\PrestaShop\helper\Data as AdyenHelperData;
+use Adyen\PrestaShop\model\Hashing;
+use Adyen\PrestaShop\service\helper\DataFactory as AdyenHelperDataFactory;
+
 require __DIR__ . '/vendor/autoload.php';
 
 if (!defined('_PS_VERSION_')) {
     exit;
 }
 
-class Adyen extends \PaymentModule
+class Adyen extends PaymentModule
 {
+    /**
+     * @var string
+     */
+    public $display;
+
+    /**
+     * @var int
+     */
+    public $is_eu_compatible;
+
+    /**
+     * @var string
+     */
+    public $meta_title;
+
+    /**
+     * @var AdyenHelperData
+     */
+    private $helper_data;
+
+    /**
+     * @var Hashing
+     */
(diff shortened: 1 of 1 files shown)


adyen.php
```diff
@@ -223,50 +259,50 @@ public function getContent()
...
+            $notification_password = (string)Tools::getValue('ADYEN_NOTI_PASSWORD');
+            $notification_hmac = (string)Tools::getValue('ADYEN_NOTI_HMAC');
+            $cron_job_token = Tools::getValue('ADYEN_CRONJOB_TOKEN');
+            $api_key_test = Tools::getValue('ADYEN_APIKEY_TEST');
+            $api_key_live = Tools::getValue('ADYEN_APIKEY_LIVE');
+            $live_endpoint_url_prefix = (string)Tools::getValue('ADYEN_LIVE_ENDPOINT_URL_PREFIX');
 
             // validating the input
-            if (!$merchant_account || empty($merchant_account) || !\Validate::isGenericName($merchant_account)) {
+            if (!$merchant_account || empty($merchant_account) || !Validate::isGenericName($merchant_account)) {
```
```suggestion
            if (empty($merchant_account) || !Validate::isGenericName($merchant_account)) {
```

If we remove the `!$merchant_account` the result will be always the same or we need to check it specifically for something?

adyen.php
```diff
@@ -223,50 +259,50 @@ public function getContent()
...
                 $output .= $this->displayError($this->l('Invalid Configuration value for Merchant Account'));
             }
 
-            if (!$notification_username || empty($notification_username) || !\Validate::isGenericName($notification_username)) {
+            if (!$notification_username || empty($notification_username) || !Validate::isGenericName($notification_username)) {
```
```suggestion
            if (empty($notification_username) || !Validate::isGenericName($notification_username)) {
```

adyen.php
```diff
@@ -223,50 +259,50 @@ public function getContent()
...
                 $output .= $this->displayError($this->l('Invalid Configuration value for Notification Username'));
             }
 
-            if (!$notification_password || empty($notification_password) || !\Validate::isGenericName($notification_password)) {
+            if (!$notification_password || empty($notification_password) || !Validate::isGenericName($notification_password)) {
```
```suggestion
            if (empty($notification_password) || !Validate::isGenericName($notification_password)) {
```

PR #23, opened 2019-10-04
[PW-1741] Create adapter classes for Prestashop

cyattilakiss:
- Create getTemplateFromModulePath helper function to retrieve the template path for both 1.6 and 1.7
- Add adapter classes for Order and CustomerThread classes

### DIFF
--- a/service/adapter/classes/order/OrderAdapter.php
+++ b/service/adapter/classes/order/OrderAdapter.php
@@ -1,0 +1,63 @@
+<?php
+
+/**
+ *                       ######
+ *                       ######
+ * ############    ####( ######  #####. ######  ############   ############
+ * #############  #####( ######  #####. ######  #############  #############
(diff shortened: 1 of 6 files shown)


service/adapter/classes/order/OrderAdapter.php
```diff
@@ -0,0 +1,63 @@
...
+ *                               ############
+ *
+ * Adyen PrestaShop module
+ *
+ * Copyright (c) 2019 Adyen B.V.
+ * This file is open source and available under the MIT license.
+ * See the LICENSE file for more info.
+ */
+
+namespace Adyen\PrestaShop\service\adapter\classes\order;
```
rikterbeek: what is the added value of added into extra folders called classes and order ?

cyattilakiss: To follow the core folder structure to find the core classes for which we have the adapter easier in the future.

PR #24, opened 2019-10-04
PW-1638: Integrate OneClick Payments

Aleffio:


controllers/front/Payment.php
```diff
@@ -400,4 +404,19 @@ public function buildBrowserData($request = [])
...
     }
 
+    /**
+     * @param array $request
+     * @param $payload
+     * @return array
+     */
+    public function buildRecurringData($request = [], $payload)
+    {
+        if (!empty($payload['storeCc']) && strpos($payload['storeCc'], 'true') !== false) {
```
cyattilakiss: why checking it with strpos? Can we do a simple === ?

helper/Data.php
```diff
@@ -100,6 +108,42 @@ public function getOriginKeyForOrigin()
...
+                ),
+            ],
+            "shopperReference" => $shopperReference,
+            "shopperLocale" => $shopperLocale
+        ];
+
+        try {
+            $responseData = $this->adyenCheckoutService->paymentMethods($adyFields);
+        } catch (\Adyen\AdyenException $e) {
+            $this->helperData->adyenLogger()->logError("There was an error retrieving the payment methods. message: " . $e->getMessage());
```
cyattilakiss: here we don't have responseData and the return will fail

views/templates/front/adyencheckout.tpl
```diff
@@ -0,0 +1,19 @@
+{if !$originKey}
+    <h5>There is an error with retrieving the originKey,
+        please check your API key in the Adyen Module configuration</h5>
+{else}
+    <script src="https://ajax.googleapis.com/ajax/libs/jquery/1.10.2/jquery.min.js"></script>
+    <script>
+        $(document).ready(function () {
+            window.adyenCheckout = new AdyenCheckout({
```
cyattilakiss: can we also pass here the paymentMethods response so we don't have to do it everyime for the other components?

views/templates/front/payment.tpl
```diff
@@ -17,8 +24,11 @@
                     <script type="text/javascript" src="/js/jquery/plugins/fancybox/jquery.fancybox.js"></script>
                 {/if}
 
+
                 <script>
 
+                    $(document).ready(function () {
+                    // window.onload = function(){
```
cyattilakiss: remove if not necessary

adyen.php
```diff
@@ -547,19 +549,64 @@ public function hookOrderConfirmation($params)
...
+        $amount = $this->context->cart->getOrderTotal();
+        $currency = $this->context->currency->iso_code;
+        $address = new Address($this->context->cart->id_address_invoice);
+        $isoAddress = Country::getIsoById($address->id_country);
+        $shopperReference = $this->context->cart->id_customer;
+        $shopperLocale = $this->context->language->iso_code;
+
+        //retrieve payment methods
+        $paymentMethods = $this->helper_data->fetchPaymentMethods($isoAddress, $amount, $currency, $shopperReference, $shopperLocale);
+        if(!empty($paymentMethods) && isset($paymentMethods['oneClickPaymentMethods'])){
```
rikterbeek: can't we simplify this to if(!empty($paymentMethods['oneClickPaymentMethods']))

adyen.php
```diff
@@ -547,19 +549,64 @@ public function hookOrderConfirmation($params)
...
+            $oneClickPaymentMethods = $paymentMethods['oneClickPaymentMethods'];
+        }
+
         $payment_options = array();
+
+        if(isset($oneClickPaymentMethods)) {
```
rikterbeek: same check as above right can't we move logic into that if statement.

adyen.php
```diff
@@ -547,19 +549,64 @@ public function hookOrderConfirmation($params)
...
+            foreach ($oneClickPaymentMethods as $storedCard) {
+                if (isset($storedCard["storedDetails"]["card"]["expiryMonth"])) {
```
rikterbeek: better to use !empty

adyen.php
```diff
@@ -547,19 +549,64 @@ public function hookOrderConfirmation($params)
      */
     public function hookPaymentOptions($params)
     {
+        $amount = $this->context->cart->getOrderTotal();
+        $currency = $this->context->currency->iso_code;
+        $address = new Address($this->context->cart->id_address_invoice);
+        $isoAddress = Country::getIsoById($address->id_country);
+        $shopperReference = $this->context->cart->id_customer;
+        $shopperLocale = $this->context->language->iso_code;
```
rikterbeek: below we use $this->helper_data->getLocale($this->context) should we use it here as well ?

adyen.php
```diff
@@ -585,19 +632,58 @@ public function hookPayment($params)
...
             array(
-                'locale' => $this->context->language->iso_code, // no locale in Prestashop1.6 only languageCode that is en-en but we need en_EN
+                'locale' => $this->helper_data->getLocale($this->context), // no locale in Prestashop1.6 only languageCode that is en-en but we need en_EN
                 'originKey' => $this->helper_data->getOriginKeyForOrigin(),
                 'environment' => \Configuration::get('ADYEN_MODE'),
                 'paymentProcessUrl' => $this->context->link->getModuleLink($this->name, 'Payment', array(), true),
                 'threeDSProcessUrl' => $this->context->link->getModuleLink($this->name, 'ThreeDSProcess', array(), true),
-                'prestashop16' => true
+                'prestashop16' => true,
+                'oneClickPaymentMethod' => "",
```
rikterbeek: why do we define an empty string  ?

views/templates/front/payment.tpl
```diff
@@ -33,12 +43,10 @@
                     let language;
                     let javaEnabled;
 
-                    let adyenCheckout;
-
                     let placeOrderAllowed;
                     let popupModal;
 
-
+                    let storeCc;
```
rikterbeek: @KadoBOT can you validate if this javascript will work for all common browsers ? I think better to move from let to var.

views/templates/front/payment.tpl
```diff
@@ -4,6 +4,13 @@
         please check your API key in the Adyen Module configuration</h5>
 </form>
 {else}
+    {if $prestashop16}
+        <p></p>
+        <div class="adyen-payment-method-label">
+            Pay with Credit Card
```
rikterbeek: Can't we add smarty translation here so merchants can change label ?

views/templates/front/adyencheckout.tpl
```diff
@@ -0,0 +1,19 @@
...
+                locale: "{$locale}",
+                originKey: "{$originKey}",
+                environment: "{$environment}",
+                risk: {
+                    enabled: false
```
rikterbeek: let's remove this and validate if it still work. This was added for a quickfix that should be solved by now.

views/templates/front/adyencheckout.tpl
Aleffio: it is on another template at the moment, will find a way to not duplicate paymentMethods response

helper/Data.php
```diff
@@ -100,6 +108,43 @@ public function getOriginKeyForOrigin()
...
+     * @param $store
+     * @param $country
+     * @return array
+     */
+    public function fetchPaymentMethods($countryCode, $amount, $currency, $shopperReference, $shopperLocale)
+    {
+        $merchantAccount = \Configuration::get('ADYEN_MERCHANT_ACCOUNT');
+
+        if (!$merchantAccount) {
+            return [];
```
cyattilakiss: would be nice to show or log a message that the merchant account is not configured therefore no payment methods are retrieved.

adyen.php
```diff
@@ -606,21 +653,55 @@ public function hookPayment()
 
         $this->context->controller->addCSS($this->_path . 'css/adyen.css', 'all');
 
+        $amount = $this->context->cart->getOrderTotal();
+        $currency = $this->context->currency->iso_code;
```
msilvagarcia: You should get the currency from the cart object:
```php
$currencyData = Currency::getCurrency($cart->id_currency);
$currency = $currencyData['iso_code'];
```

adyen.php
```diff
@@ -569,25 +570,71 @@ public function hookOrderConfirmation()
     public function hookPaymentOptions()
     {
         $payment_options = array();
+        if(!$this->context->customer->is_guest) {
+            $amount = $this->context->cart->getOrderTotal();
+            $currency = $this->context->currency->iso_code;
```
msilvagarcia: You should get the currency from the cart object:
```php
$currencyData = Currency::getCurrency($cart->id_currency);
$currency = $currencyData['iso_code'];
```

adyen.php
```diff
@@ -569,25 +570,71 @@ public function hookOrderConfirmation()
...
+            $shopperLocale = $this->helper_data->getLocale($this->context);
+
+
+            //retrieve payment methods
+            $paymentMethods = $this->helper_data->fetchPaymentMethods($isoAddress, $amount, $currency, $shopperReference, $shopperLocale);
+
+            if (!empty($paymentMethods['oneClickPaymentMethods'])) {
+                $oneClickPaymentMethods = $paymentMethods['oneClickPaymentMethods'];
+                foreach ($oneClickPaymentMethods as $storedCard) {
+                    if (!empty($storedCard["storedDetails"]["card"]["expiryMonth"])) {
```
msilvagarcia: Why is this checking for expiry month? Can we make the reason more obvious somehow?

adyen.php
```diff
@@ -606,21 +653,55 @@ public function hookPayment()
...
+                            'oneClickPaymentMethod' => json_encode($storedCard),
+                            'recurringDetailReference' => $storedCard['recurringDetailReference'],
+                            'name' => $storedCard['name'],
+                            'number' => $storedCard['storedDetails']['card']['number']
+                        )
+                    );
+                }
+                $payments .= $this->display(__FILE__, '/views/templates/front/oneclick.tpl');
+            }
+        }
```
msilvagarcia: This block doesn't look too different from the previous one. Can we extract a method to unify these?

controllers/front/Payment.php
```diff
@@ -393,4 +397,19 @@ public function buildBrowserData($request = [])
         return $request;
     }
 
+    /**
+     * @param array $request
+     * @param $payload
+     * @return array
+     */
+    public function buildRecurringData($request = [], $payload)
```
msilvagarcia: Short array syntax won't work in PHP 5.2

helper/Data.php
```diff
@@ -100,6 +108,43 @@ public function getOriginKeyForOrigin()
         return $originKey;
     }
 
+    /**
+     * @param $store
+     * @param $country
+     * @return array
+     */
+    public function fetchPaymentMethods($countryCode, $amount, $currency, $shopperReference, $shopperLocale)
```
msilvagarcia: Most of these parameters are in the Cart object. Passing the Cart object here would make calling this object less complex.

helper/Data.php
```diff
@@ -100,6 +108,43 @@ public function getOriginKeyForOrigin()
...
+        ];
+
+        $responseData = "";
+        try {
+            $responseData = $this->adyenCheckoutService->paymentMethods($adyFields);
+        } catch (\Adyen\AdyenException $e) {
+            $this->helperData->adyenLogger()->logError("There was an error retrieving the payment methods. message: " . $e->getMessage());
+        }
+        return $responseData;
+    }
```
msilvagarcia: This method would be better in a specialized object

helper/Data.php
```diff
@@ -375,6 +420,20 @@ public function startSession()
...
+    public function getLocale($context)
+    {
+        // no locale in Prestashop1.6 only languageCode that is en-en but we need en_EN
+        if ($this->isPrestashop16()) {
+            return $context->language->iso_code;
+        } else {
+            return $context->language->locale;
+        }
+    }
+
```
msilvagarcia: This method would be better in a specialized object

views/templates/front/payment.tpl
msilvagarcia: https://caniuse.com/#feat=let
Doesn't work until IE 10

controllers/front/Payment.php
```diff
@@ -286,6 +287,17 @@ public function buildCCData($request, $payload)
...
 
+        $shopperReference = $this->context->cart->id_customer;
+        if(!empty($shopperReference)) {
+            $request['shopperReference'] = $shopperReference;
+        }
+
+        //Oneclick data
+        if(!empty($payload['recurringDetailReference']) &&
+        $recurringDetailReference = $payload['recurringDetailReference']) {
+            $request['paymentMethod']['recurringDetailReference'] = $recurringDetailReference;
```
cyattilakiss:
```suggestion
            $request['paymentMethod']['recurringDetailReference'] = $payload['recurringDetailReference'];
```

Bit easier to read IMO but you can keep as it is if you think :)

controllers/front/Payment.php
```diff
@@ -286,6 +287,17 @@ public function buildCCData($request, $payload)
...
         }
 
+        $shopperReference = $this->context->cart->id_customer;
+        if(!empty($shopperReference)) {
+            $request['shopperReference'] = $shopperReference;
+        }
+
+        //Oneclick data
+        if(!empty($payload['recurringDetailReference']) &&
+        $recurringDetailReference = $payload['recurringDetailReference']) {
```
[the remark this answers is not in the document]
cyattilakiss: or maybe the whole recheck is not even important since the !empty part will check these so the `&&
        $recurringDetailReference = $payload['recurringDetailReference']` can just be removed

PR #27, opened 2019-10-08
PW-1637

msilvagarcia:
Add a refund process to the PrestaShop plugins

### DIFF
--- a/upgrade/upgrade-1.1.0.php
+++ b/upgrade/upgrade-1.1.0.php
@@ -34,5 +34,7 @@ if (!defined('_PS_VERSION_')) {
  */
 function upgrade_module_1_1_0(Adyen $module)
 {
-    return $module->registerHook('actionOrderSlipAdd');
+   $module->registerHook('actionOrderSlipAdd');
+   $module->registerHook('displayPaymentTop');
+   return true;
 }


service/modification/Refund.php
```diff
@@ -0,0 +1,108 @@
...
+from %sadyen_notification a
+inner join %sorders o on a.merchant_reference = o.id_cart
+where o.id_order = $orderId
+SQL
+            ,
+            _DB_PREFIX_, _DB_PREFIX_
+        ));
+        $this->logger->logDebug($results);
+        $this->modificationClient->refund([
+            'originalReference' => $results[0]['pspReference'],
```
rikterbeek: shall we defence program this and check if results has a value and $results[0]['pspReference'] exists ?

service/modification/Refund.php
```diff
@@ -0,0 +1,108 @@
...
+            <<<SQL
+select a.pspreference as pspReference,
+   o.id_order as orderId
+from %sadyen_notification a
+inner join %sorders o on a.merchant_reference = o.id_cart
+where o.id_order = $orderId
+SQL
+            ,
+            _DB_PREFIX_, _DB_PREFIX_
+        ));
```
rikterbeek: shall we move all database query to it's own method so it is clear what the query does ?

service/adapter/classes/order/OrderAdapter.php
```diff
@@ -14,7 +14,7 @@
  *                               #############
  *                               ############
  *
- * Adyen PrestaShop module
+ * Adyen PrestaShop plugin
```
cyattilakiss: We are going to rename all the plugin to module in a later step so that is why you see my new files with the module and not with the plugin :)
In the moment doesn't matter too much

service/modification/Refund.php
```diff
@@ -0,0 +1,108 @@
...
+    ) {
+        $this->modificationClient = $modificationClient;
+        $this->db = $db;
+        $this->merchantAccount = $merchantAccount;
+        $this->logger = $logger;
+    }
+
+    /**
+     * @param string $orderId
+     * @param string $amount
```
Aleffio:
```suggestion
     * @param OrderSlip $orderSlip
```

service/modification/Refund.php
```diff
@@ -0,0 +1,108 @@
...
+     * @return bool
+     * @throws AdyenException
+     * @throws PrestaShopDatabaseException
+     */
+    public function request($orderId, OrderSlip $orderSlip, $currency)
+    {
+        $amount = (new Currency())->sanitize($orderSlip->amount, $currency);
+
+        $orderId = pSQL($orderId);
+        $results = $this->db->executeS(sprintf(
```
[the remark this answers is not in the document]
msilvagarcia: Because multi-line strings are a pain to deal with, so using a Heredoc makes the query itself easier to read while keeping the file structure somewhat sane. And then Heredocs don't support concatenation, so `sprintf` deals with that.

upgrade/upgrade-1.1.0.php
```diff
@@ -0,0 +1,38 @@
...
+ * This function is automatically called on version upgrades.
+ *
+ * Version 1.1.0 introduces refund feature
+ *
+ * @param Adyen $module
+ * @return bool
+ */
+function upgrade_module_1_1_0(Adyen $module)
+{
+    return $module->registerHook('actionOrderSlipAdd');
```
Aleffio:
```suggestion
   $module->registerHook('actionOrderSlipAdd');
   $module->registerHook('displayPaymentTop');
   return true;
```

rikterbeek:
@msilvagarcia can you solve the conflict ?

service/modification/Refund.php
rikterbeek: @cyattilakiss where do you see this check ?

upgrade/upgrade-1.1.0.php
msilvagarcia: I believe the best way to go about this is for you to add it to your repository as version 1.1.1 or something

msilvagarcia:
@rikterbeek done :)

PR #29, opened 2019-10-15
Add PHP Code Sniffer

msilvagarcia:


adyen.php
```diff
@@ -20,6 +20,8 @@
  * See the LICENSE file for more info.
  */
 
+// phpcs:disable PSR1.Classes.ClassDeclaration
+
```
rikterbeek: What does this do exactly ?

controllers/admin/AdminAdyenPrestashopCronController.php
```diff
@@ -13,13 +13,15 @@
  *                               #############
  *                               ############
  *
- * Adyen PrestaShop module
+ * Adyen PrestaShop plugin
```
rikterbeek: Prestashop uses the temonology modules https://addons.prestashop.com/en/ shall we call it a module ? ask Grant to verify this.

controllers/front/Notifications.php
```diff
@@ -20,6 +20,8 @@
  * See the LICENSE file for more info.
  */
 
+// phpcs:disable PSR1.Classes.ClassDeclaration
+
```
rikterbeek: same here ?

adyen.php
msilvagarcia: Disables warning about a class not being part of a namespace.

PR #32, opened 2019-10-25
PW-1761 Display Local Payment Methods

msilvagarcia:
This pull request will show a list of available Local Payment Methods


adyen.php
```diff
@@ -619,6 +614,51 @@ public function hookPaymentOptions()
             }
         }
 
+        if (!empty($paymentMethods['paymentMethods'])) {
+            foreach ($paymentMethods['paymentMethods'] as $paymentMethod) {
+                $issuerList = [];
+                if (empty($paymentMethod['type']) || $paymentMethod['type'] == 'scheme') {
+                    continue;
+                }
+                if (isset($paymentMethod['details'])) {
```
```suggestion
                if (!empty($paymentMethod['details'])) {
```

So if it is an empty array or not set then we won't even try to foreach through it.

adyen.php
```diff
@@ -684,6 +724,42 @@ public function hookPayment()
             }
         }
 
+        if (!empty($paymentMethods['paymentMethods'])) {
```
cyattilakiss: Does it worth to follow WET and not creating a helper function which prepares the issuerlist ?

adyen.php
msilvagarcia: In this case, I don't assume the performance impact is going to matter. I assume the compiler will just skip this if it's an empty array.

We can enforce some similar conditions by adding a check for `is_array`. I just prefer to make the condition obvious instead of abstract.

adyen.php
msilvagarcia: What's WET?

adyen.php
rikterbeek: I would prefer !empty() over isset in this case as it just do an extra check if it has a value. is_array is indeed a bit overkill

adyen.php
```diff
@@ -684,6 +724,42 @@ public function hookPayment()
...
+                            array(),
+                            true
+                        ),
+                        'renderPayButton' => true,
+                    )
+                );
+                $payments .= $this->display(__FILE__, '/views/templates/front/local-payment-method.tpl');
+            }
+        }
+
```
rikterbeek: it's a huge method can we break it into several methods ?

rikterbeek:
@msilvagarcia can you rebase the branch ?

PR #35, opened 2019-11-01
PW-1762 Redirect shopper to external authorization

msilvagarcia:


adyen.php
```diff
@@ -975,4 +910,109 @@ private function createCustomerMessage($message, CustomerThread $customerThread)
...
+    /**
+     * @return string
+     */
+    private function getStandardPaymentMethod()
+    {
+        $payments = '';
+        $this->context->smarty->assign(
+            array(
+                'locale' => $this->helper_data->getLocale($this->context->language),
+                // no locale in PrestaShop1.6 only languageCode that is en-en but we need en_EN
```
rikterbeek: should we define a default ?

msilvagarcia: what do you mean with this?

controllers/front/Payment.php
```diff
@@ -412,4 +421,94 @@ public function buildRecurringData($request = array(), $payload)
...
+    }
+
+    private function processLocalPaymentMethod(\Cart $cart, $paymentType, $paymentIssuer)
+    {
+        $request = array();
+        $request = $this->buildBrowserData($request);
+        $request = $this->buildLocalPaymentMethodData($paymentType, $paymentIssuer, $request);
+        $request = $this->buildPaymentData($request);
+        $request = $this->buildMerchantAccountData($request);
+        $request = $this->buildRecurringData($request, $_REQUEST);
```
Aleffio: I think we don't need this one

controllers/front/Payment.php
```diff
@@ -412,4 +421,94 @@ public function buildRecurringData($request = array(), $payload)
...
+                    [
+                        'message' => "There was an error with the payment method, please choose another one."
+                    ]
+                )
+            );
+            return;
+        }
+
+        switch ($response['resultCode']) {
+            case 'RedirectShopper':
```
Aleffio: Do we need a Default case to non-handle(error) the other flows? (I can only think of PresentToShopper at the moment)

msilvagarcia: I will add that

controllers/front/Payment.php
msilvagarcia: We do. The `$request` array is just created there.

adyen.php
rikterbeek: You are specifying that for 1.6 the format is incorrect en-en instaed of en_EN should we then not prefine it for 1.6 to en_EN or should we make a converter that makes en-en to en_EN ?

Aleffio: I think we need to do this in the getLocale function in the helper_data, I will create a youtrack ticket for this, so we can discuss

controllers/front/Payment.php
Aleffio: buildRecurringData fills the request only if 'storeCc' is true, which is the checkbox we use for saving the credit cards, we don't need this for local payment methods. I am referring only to the buildRecurringData, the others are needed oc.

msilvagarcia:
```suggestion
```

PR #41, opened 2019-11-13
Fix payment method check

msilvagarcia:


adyen.php
```diff
@@ -1047,11 +1047,15 @@ private function getStandardPaymentMethod()
      */
     private function isSimplePaymentMethod($paymentMethod)
     {
+        $details = $paymentMethod['details'];
```
```suggestion
        $details = false;
        if (!empty($paymentMethod['details'])) {
            $details = $paymentMethod['details'];
        }
```

adyen.php
```diff
@@ -1047,11 +1047,15 @@ private function getStandardPaymentMethod()
...
         return !empty($paymentMethod['type'])
             && $paymentMethod['type'] != 'scheme'
             && (
-                empty($paymentMethod['details']) || (
-                    $paymentMethod['details'][0]['key'] == 'issuer' && $paymentMethod['details'][0]['type'] == 'select'
+                empty($details) || (
```
```suggestion
```

adyen.php
```diff
@@ -1047,11 +1047,15 @@ private function getStandardPaymentMethod()
...
+                    is_array($details)
+                    && count($details) == 1
+                    && $details[0]['key'] == 'issuer'
+                    && $details[0]['type'] == 'select'
                 )
```
```suggestion
```

adyen.php
```diff
@@ -1047,11 +1047,15 @@ private function getStandardPaymentMethod()
...
+        $details = $paymentMethod['details'];
         return !empty($paymentMethod['type'])
             && $paymentMethod['type'] != 'scheme'
             && (
-                empty($paymentMethod['details']) || (
-                    $paymentMethod['details'][0]['key'] == 'issuer' && $paymentMethod['details'][0]['type'] == 'select'
+                empty($details) || (
+                    is_array($details)
+                    && count($details) == 1
+                    && $details[0]['key'] == 'issuer'
```
cyattilakiss: It can happen that the issuer is not the key of the [0] element of the array I thin we should follow this approach how we did it on the frontend in M2: https://github.com/Adyen/adyen-magento2/blob/5.0.1/view/frontend/web/js/view/payment/method-renderer/adyen-hpp-method.js#L282

PR #42, opened 2019-11-18
PW-1799: Refactor the plugin front-end code

msilvagarcia:


adyen.php
```diff
@@ -1061,4 +1022,105 @@ private function isSimplePaymentMethod($paymentMethod)
...
+     * @throws \PrestaShop\PrestaShop\Adapter\CoreException
+     */
+    private function registerAdyenJavascript($controller)
+    {
+        /** @var \Adyen\PrestaShop\service\adapter\classes\Controller $controllerAdapter */
+        $controllerAdapter = $this->getService('Adyen\PrestaShop\service\adapter\classes\Controller');
+        $controllerAdapter->setController($controller);
+        if ($this->helper_data->isDemoMode()) {
+            $controllerAdapter->registerJavascript(
+                'adyen-component', // Unique ID
```
cyattilakiss: if this should be a unique ID I would say use: 
```suggestion
                'adyen-checkout-component', // Unique ID
```

adyen.php
```diff
@@ -1061,4 +1022,105 @@ private function isSimplePaymentMethod($paymentMethod)
...
+                \Adyen\PrestaShop\service\Configuration::CHECKOUT_COMPONENT_JS_TEST, // JS path
+                array('server' => 'remote', 'position' => 'bottom', 'priority' => 150) // Arguments
+            );
+            $controllerAdapter->registerJavascript(
+                'adyen-threeDS2Utils', // Unique ID
+                $this->_path . 'views/js/threeds2-js-utils.js', // JS path
+                array('position' => 'bottom', 'priority' => 160) // Arguments
+            );
+            $controllerAdapter->registerStylesheet(
+                'adyen-stylecheckout', // Unique ID
```
cyattilakiss: would use the same uniq id:
IF the css and js has these uniq ids in a different pool
```suggestion
                'adyen-checkout-component', // Unique ID
```

adyen.php
```diff
@@ -679,7 +635,7 @@ public function hookPaymentOptions()
                 'paymentProcessUrl' => $this->context->link->getModuleLink($this->name, 'Payment', array(), true),
                 'threeDSProcessUrl' => $this->context->link->getModuleLink($this->name, 'ThreeDSProcess', array(), true),
                 'prestashop16' => false,
-                'loggedInUser' => !$this->context->customer->is_guest
+                'loggedInUser' => (int)!$this->context->customer->is_guest
```
rikterbeek: is this always a integer ? I could imagine this default value can be overwritten to be any value.

msilvagarcia: The value here would be a boolean

rikterbeek: clear.

PR #46, opened 2019-11-22
[PW-1883] Fix reset functionality for module

cyattilakiss:
Add reset function to handle reset only the parameters option
Fix reset "reset everything" option
Fix decrypt php error if parameter is empty

### DIFF
--- a/adyen.php
+++ b/adyen.php
@@ -110,9 +110,15 @@ class Adyen extends PaymentModule
         $this->confirmUninstall = $this->l('Are you sure you want to uninstall?');
     }
 
-    /*
-	 * for installing the plugin
-	 */
+    /**
+     * Install script
+     *
+     * This function is called when
+     * when User installs the module or
+     * when User resets the module and selects the do not keep the data option
+     *
+     * @return bool
+     */
     public function install()
     {
         if (!$this->versionChecker->isPrestaShopSupportedVersion()) {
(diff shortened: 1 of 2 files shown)


helper/Data.php
```diff
@@ -187,6 +187,10 @@ public function encrypt($data)
 
     public function decrypt($data)
     {
+        if (empty($data)) {
+            return '';
```
rikterbeek: when should this be empty ? should we log this ?

adyen.php
```diff
@@ -236,9 +276,15 @@ private function removeConfigurationsFromDatabase()
...
         );
 
-        $db = Db::getInstance();
-        /** @noinspection SqlWithoutWhere SqlResolve */
-        return $db->execute('DELETE FROM `' . _DB_PREFIX_ . 'configuration` WHERE `name` IN ("' . implode('", "', $adyenConfigurationNames) . '") ');
+        $result = true;
+
+        foreach ($adyenConfigurationNames as $adyenConfigurationName) {
+            if (!Configuration::deleteByName($adyenConfigurationName)) {
+                $result = false;
```
rikterbeek: should we log this as this should not happen I believe.

helper/Data.php
cyattilakiss: If something goes wrong I believe - last time it was empty because the cron job token was empty. Yes we can log this, could help with debugging.

adyen.php
cyattilakiss: Yes we can log this as well. Only happens if default prestashop functionality breaks, so something unexpected.

adyen.php
```diff
@@ -236,9 +276,16 @@ private function removeConfigurationsFromDatabase()
...
         );
 
-        $db = Db::getInstance();
-        /** @noinspection SqlWithoutWhere SqlResolve */
-        return $db->execute('DELETE FROM `' . _DB_PREFIX_ . 'configuration` WHERE `name` IN ("' . implode('", "', $adyenConfigurationNames) . '") ');
+        $result = true;
+
+        foreach ($adyenConfigurationNames as $adyenConfigurationName) {
+            if (!Configuration::deleteByName($adyenConfigurationName)) {
+                $this->adyenLogger()->logDebug("Configuration couldn't be deleted by name: " . $adyenConfigurationName);
```
Aleffio:
```suggestion
                $this->helper_data->adyenLogger()->logDebug("Configuration couldn't be deleted by name: " . $adyenConfigurationName);
```

PR #49, opened 2019-12-09
[PW-1810] Update checkout component version from 3.0.0 to 3.4.0

cyattilakiss:
- Bump checkout component version from 3.0.0 to 3.4.0
- Use component's state.date.browserInfo instead of the external
browserInfo script
- Remove threeds2-js-utils.js
- Rename recurringDetailReference to storedPaymentMethodId
- Use paymentMethods response storedPaymentMethods instead of the
oneClickPaymentMethods object
- Use the paymentMethods response provided back from the componenent
on the frontend instead of providing the response directly via smarty
- Use polyfill for Object.assign browser compatibility


views/js/payment-components/one-click.js
```diff
@@ -41,22 +40,31 @@ jQuery(function ($) {
...
+    // Iterate through the stored payment methods list we got from the adyen checkout component
+    checkoutStoredPaymentMethods.forEach(function (storedPaymentMethod) {
+
+        /*
+            storedPaymentMethod.id = $storedPaymentApiId in the oneclick.tpl
+            if the container doesn't exits don't try to render the component
+        */
+        var storedPaymentMethodContainer = $("#cardContainer-" + storedPaymentMethod.id);
+        if (!storedPaymentMethodContainer.length) {
+            console.log("Something went wrong with the payment method");
```
msilvagarcia: I wouldn't log this here. Logging like this can be seen by shoppers.

views/js/payment-components/one-click.js
```diff
@@ -80,32 +88,35 @@ jQuery(function ($) {
...
      */
-    function renderOneClickComponent(oneClickPaymentMethod) {
-        var card = window.adyenCheckout.create('card', {
-            type: oneClickPaymentMethod.type,
-            oneClick: true,
-            details: oneClickPaymentMethod.details,
-            storedDetails: oneClickPaymentMethod.storedDetails,
-
+    function renderStoredPaymentComponent(storedPaymentMethod) {
+        var card = window.adyenCheckout.create('card', Object.assign(
```
msilvagarcia: `Object.assign` is not available in any version of Internet Explorer

views/js/payment-components/one-click.js
```diff
@@ -41,22 +40,31 @@ jQuery(function ($) {
...
+            if the container doesn't exits don't try to render the component
+        */
+        var storedPaymentMethodContainer = $("#cardContainer-" + storedPaymentMethod.id);
+        if (!storedPaymentMethodContainer.length) {
+            console.log("Something went wrong with the payment method");
+            return;
         }
 
-        threeDSProcessUrl = oneClickPaymentConfiguration.threeDsProcessUrl;
+        var storedPaymentMethodConfiguration = $('[data-stored-payment-api-id="' + storedPaymentMethod.id + '"]').data();
```
msilvagarcia:
```suggestion
        var storedPaymentMethodConfiguration = storedPaymentMethodContainer.find('[data-stored-payment-api-id="' + storedPaymentMethod.id + '"]').data();
```
I don't know if this is possible, but would be better to restrict the search to the container

views/js/payment-components/one-click.js
```diff
@@ -41,22 +40,31 @@ jQuery(function ($) {
...
-        var oneClickPaymentConfiguration = $(element).data();
-        var oneClickPaymentMethod = oneClickPaymentConfiguration.oneClickPayment;
-        if (oneClickPaymentMethod) {
-            renderOneClickComponent(oneClickPaymentMethod);
-            fillBrowserInfo();
+    // use this object to iterate through the stored payment methods
+    var checkoutStoredPaymentMethods = window.adyenCheckout.paymentMethodsResponse.storedPaymentMethods;
+
+    // Iterate through the stored payment methods list we got from the adyen checkout component
+    checkoutStoredPaymentMethods.forEach(function (storedPaymentMethod) {
```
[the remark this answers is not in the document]
cyattilakiss: Checked with checkout. 10 11 Edge.

views/js/payment-components/one-click.js
cyattilakiss: unfortunately the div that we want to find is outside of the storedPaymentMethodContainer so we can relay on the storedPaymentMethod.id

views/js/payment-components/one-click.js
cyattilakiss: indeed, I will have to find a solution for this one

views/js/payment-components/one-click.js
cyattilakiss: ok

views/js/payment-components/stored-payment-method.js
```diff
@@ -41,22 +40,32 @@ jQuery(function ($) {
...
+        */
+        var storedPaymentMethodContainer = $("#cardContainer-" + storedPaymentMethod.id);
+
+        // container doesn't exist, something went wrong on the template side
+        if (!storedPaymentMethodContainer.length) {
+            return;
         }
 
-        threeDSProcessUrl = oneClickPaymentConfiguration.threeDsProcessUrl;
+        var storedPaymentMethodConfiguration = $('[data-stored-payment-api-id="' + storedPaymentMethod.id + '"]').data();
```
rikterbeek: can't we better do storedPaymentMethod.data() ?

views/js/payment-components/stored-payment-method.js
```diff
@@ -80,32 +89,53 @@ jQuery(function ($) {
...
+     */
+    function mergeObjects(objectArray)
+    {
+        return objectArray.reduce(function (r, o) {
+            Object.keys(o).forEach(function (k) {
+                r[k] = o[k];
+            });
+            return r;
+        }, {});
     }
```
rikterbeek: is there not a nicer way then this ?

views/js/payment-components/stored-payment-method.js
msilvagarcia: I checked with Attila as well. The elements are different and not in the same scope, so that's not possible without changing the HTML. I agree with your comment, though.

views/js/payment-components/credit-card.js
```diff
@@ -106,6 +105,13 @@ jQuery(function ($) {
                         encryptedSecurityCode = state.data.paymentMethod.encryptedSecurityCode;
                     }
 
+                    javaEnabled = state.data.browserInfo.javaEnabled;
+                    colorDepth = state.data.browserInfo.colorDepth;
+                    screenWidth = state.data.browserInfo.screenWidth;
+                    screenHeight = state.data.browserInfo.screenHeight;
+                    timeZoneOffset = state.data.browserInfo.timeZoneOffset;
+                    language = state.data.browserInfo.language;
+
```
KadoBOT: Can't you use the `state.data.browserInfo` instead assigning the values to variables with the same name?

KadoBOT: So, instead of:
```
            processPayment({
                'isAjax': true,
                'holderName': holderName,
                'encryptedCardNumber': encryptedCardNumber,
                'encryptedExpiryMonth': encryptedExpiryMonth,
                'encryptedExpiryYear': encryptedExpiryYear,
                'encryptedSecurityCode': encryptedSecurityCode,
                'storeCc': storeCc,
                'browserInfo': {
                    'screenWidth': screenWidth,
                    'screenHeight': screenHeight,
                    'colorDepth': colorDepth,
                    'timeZoneOffset': timeZoneOffset,
                    'language': language,
                    'javaEnabled': javaEnabled
                }
            });
```

Will be:

```
            processPayment({
                'isAjax': true,
                'holderName': holderName,
                'encryptedCardNumber': encryptedCardNumber,
                'encryptedExpiryMonth': encryptedExpiryMonth,
                'encryptedExpiryYear': encryptedExpiryYear,
                'encryptedSecurityCode': encryptedSecurityCode,
                'storeCc': storeCc,
                'browserInfo': browserInfo
            });
```

views/js/payment-components/stored-payment-method.js
```diff
@@ -41,22 +40,32 @@ jQuery(function ($) {
...
-        var oneClickPaymentConfiguration = $(element).data();
-        var oneClickPaymentMethod = oneClickPaymentConfiguration.oneClickPayment;
-        if (oneClickPaymentMethod) {
-            renderOneClickComponent(oneClickPaymentMethod);
-            fillBrowserInfo();
+    // use this object to iterate through the stored payment methods
+    var checkoutStoredPaymentMethods = window.adyenCheckout.paymentMethodsResponse.storedPaymentMethods;
+
+    // Iterate through the stored payment methods list we got from the adyen checkout component
+    checkoutStoredPaymentMethods.forEach(function (storedPaymentMethod) {
```
KadoBOT: `forEach` method is not supported in IE11 and below

KadoBOT: You can use jQuery `each`. Ie:
```
$.each([ 52, 97 ], function( index, value ) {
  alert( index + ": " + value );
});
```

KadoBOT: Or for loop 🤮

views/js/payment-components/stored-payment-method.js
```diff
@@ -41,22 +40,32 @@ jQuery(function ($) {
...
+
+        /*
+            storedPaymentMethod.id = $storedPaymentApiId in the stored-payment-method.tpl
+            if the container doesn't exits don't try to render the component
+        */
+        var storedPaymentMethodContainer = $("#cardContainer-" + storedPaymentMethod.id);
+
+        // container doesn't exist, something went wrong on the template side
+        if (!storedPaymentMethodContainer.length) {
+            return;
```
KadoBOT: Instead of returning void, and negating the condition, can't you put the logic inside the if statement?
```javascript
if (storedPaymentMethodContainer.length) {
// logic here
} // END
```

views/js/payment-components/stored-payment-method.js
```diff
@@ -80,32 +89,53 @@ jQuery(function ($) {
...
+    }
+
+    /**
+     * Merge objects from an array of objects, key by key, IE 9+ compatible
+     * @param objectArray
+     * @returns {*}
+     */
+    function mergeObjects(objectArray)
+    {
+        return objectArray.reduce(function (r, o) {
```
KadoBOT: `reduce` and `forEach` IE11 🙈

views/js/payment-components/stored-payment-method.js
KadoBOT: This part of the code is confusing. I'm pretty sure there is a better way... also `r`, `o`, `k` ? ¯\_(ツ)_/¯

views/js/payment-components/stored-payment-method.js
msilvagarcia: We can, but we usually prefer returning earlier. Is there anything other than code style that we're missing in this situation?

KadoBOT: You don't need to return anything after the `if` the `if` condition is not met

PR #50, opened 2019-12-10
[PW-1893] Implement Multibanco

cyattilakiss:


controllers/front/Payment.php
```diff
@@ -514,6 +509,26 @@ private function processLocalPaymentMethod(\Cart $cart, $paymentType, $paymentIs
                     // invalid parameters
                     $isValid = false;
                 }
+                break;
+            case 'PresentToShopper':
+
+                $_SESSION['paymentAction'] = $response['action'];
+
+                $customer = new \Customer($cart->id_customer);
```
rikterbeek: should we not validate if $cart->id_customer has a value before loading it into the customer object ?

views/js/payment-components/local-payment-method.js
```diff
@@ -24,23 +24,53 @@ jQuery(function ($) {
...
+     */
+    function mergeObjects(objectArray)
+    {
+        return objectArray.reduce(function (r, o) {
+            Object.keys(o).forEach(function (k) {
+                r[k] = o[k];
+            });
+            return r;
+        }, {});
+    }
```
rikterbeek: no other way possible ?

views/js/payment-components/stored-payment-method.js
```diff
@@ -46,10 +46,8 @@ jQuery(function ($) {
     // Iterate through the stored payment methods list we got from the adyen checkout component
     checkoutStoredPaymentMethods.forEach(function (storedPaymentMethod) {
 
-        /*
-            storedPaymentMethod.id = $storedPaymentApiId in the stored-payment-method.tpl
-            if the container doesn't exits don't try to render the component
-        */
+        //  storedPaymentMethod.id = $storedPaymentApiId in the stored-payment-method.tpl
+        //  if the container doesn't exits don't try to render the component
```
rikterbeek: remove this or make it the comment clear.

controllers/front/Payment.php
cyattilakiss: prestashop will add an id_customer even if it is a guest user. If by any chance there is an issue with the customer id the next line: Validate::isLoadedObject($customer) will validate it.

views/js/payment-components/stored-payment-method.js
cyattilakiss: removed, not necessary

views/js/payment-components/local-payment-method.js
cyattilakiss: because of IE compatibility unfortunately not.

adyen.php
```diff
@@ -781,6 +776,20 @@ public function hookDisplayPaymentEU()
         return $payment_options;
     }
 
+    /**
+     * Retrieve the necessary default configuration parameters for a checkout component
+     *
+     * @return array
+     */
+    public function getCheckoutComponentInitData()
```
msilvagarcia: Can this method be private instead?

PR #53, opened 2019-12-17
PW-1891 Treat a `Refused` error code for 3DS process

msilvagarcia:
As of now, our PrestaShop plugin would show a blank page with a JSON with the error message. This is because the back-end would send an empty response with a 200 status code, tricking the front-end into treating this as a successful payment.

Changing the action of this response to `error` will inform the front-end how to properly handle this issue.

### DIFF
--- a/controllers/front/ThreeDSProcess.php
+++ b/controllers/front/ThreeDSProcess.php
@@ -118,6 +118,15 @@ class AdyenThreeDSProcessModuleFrontController extends \Adyen\PrestaShop\control
             ));
         }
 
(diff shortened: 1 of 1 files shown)


controllers/front/ThreeDSProcess.php
```diff
@@ -118,6 +118,15 @@ public function postProcess()
             ));
         }
 
+        if ($result['resultCode'] == 'Refused') {
```
Aleffio:
```suggestion
        if (!empty($result['resultCode']) &&
            $result['resultCode'] === 'Refused') {
```

PR #56, opened 2020-01-06
Add index.php files to directories

msilvagarcia:
According to the PrestaShop Code Validator, having an index.php file in each directory helps with the e-commerce security.

These files were added with the [autoindex](https://github.com/jmcollin/autoindex) package.

### DIFF
--- a/.github/index.php
+++ b/.github/index.php
@@ -1,0 +1,31 @@
+<?php
+/**
+ *                       ######
+ *                       ######
+ * ############    ####( ######  #####. ######  ############   ############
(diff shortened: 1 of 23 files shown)


rikterbeek:
@msilvagarcia shall we add this as a devDependency in the composer file so it is easily to use for us all ?

msilvagarcia:
@rikterbeek We need to figure a way to set the file with our copyright header to do so. The app uses the file index.php that's inside the sources folder in itself, so either fork it with our version of the index.php or add the feature to configure the index.php to copy from.

PR #58, opened 2020-01-08
[PW-1809] Improve logging in Prestashop

cyattilakiss:
Add new exception: CommandException
Add and configure new logger based on Monolog
Add new logger to Adyen PHP library client as a default logger 
Depricate adyenLogger FileLogger in the helper\Data class
Replace depricated logger with the new one


service/logger/Logger.php
```diff
@@ -0,0 +1,197 @@
...
+ * See the LICENSE file for more info.
+ */
+
+namespace Adyen\PrestaShop\service\logger;
+
+use Adyen\PrestaShop\application\VersionChecker;
+use Monolog\Handler\StreamHandler;
+use Adyen\PrestaShop\exception\CommandException;
+
+class Logger extends \Monolog\Logger
```
cyattilakiss: make it final

service/logger/Logger.php
```diff
@@ -0,0 +1,197 @@
...
+        if (!file_exists($path)) {
+            if (!mkdir($path, 0755, true)) {
+                throw new CommandException('Creating the Adyen log folder failed');
+            }
+        }
+
+        return $path;
+    }
+
+    public function addAdyenAPI($message, array $context = [])
```
cyattilakiss: phpdocs

rikterbeek:
@cyattilakiss WIP ? let me know when we can review it.

adyen.php
```diff
@@ -1136,6 +1144,7 @@ private function getStandardPaymentMethod()
     private function isUnsupportedPaymentMethod($paymentMethodType)
     {
         $unsupportedPaymentMethods = array(
+            'scheme',
```
msilvagarcia: Is this on purpose?

controllers/admin/AdminAdyenPrestashopCronController.php
```diff
@@ -41,6 +41,7 @@ class AdminAdyenPrestashopCronController extends \ModuleAdminController
     public function __construct()
     {
         $this->helperData = ServiceLocator::get('Adyen\PrestaShop\helper\Data');
+        $this->logger = ServiceLocator::get('Adyen\PrestaShop\service\Logger');
```
msilvagarcia: This property is not declared in the class

helper/Data.php
```diff
@@ -162,6 +179,10 @@ public function isDemoMode()
         }
     }
 
+    /**
+     * @deprecated Use \Adyen\PrestaShop\service\logger instead
```
msilvagarcia:
```suggestion
     * @deprecated Use \Adyen\PrestaShop\service\logger instead. This method will be removed in version 2.
```

adyen.php
cyattilakiss: yes - but wrong branch
I will create another branch to fix some issues - I just noticed that the card payments are included anyway and doesn't check the paymentMethods response so we show the scheme 2 times if it is in the paymentMethods list which was never noticed because the isSimplePaymentMethod function filtered the original one. :(

controllers/admin/AdminAdyenPrestashopCronController.php
cyattilakiss: indeed I will add the helperData and the logger as a property and not just dynamically declaring them

cyattilakiss:
> Please, take a look at the comments I added inline.
> 
> I also wonder if you could have used the LoggerInterface more often.

In the `ServiceLocator::get()` you mean?

PR #59, opened 2020-01-16
[PW-1730] OpenInvoice payment method support

cyattilakiss:
Invoicing and delivery address is parsed assuming the followings:
Address line 1 is labeled as "street name"
Address line 2 is labeled as "house number"


controllers/front/Payment.php
```diff
@@ -440,20 +482,65 @@ private function buildReturnURL($request)
...
+        $request = $this->customerBuilder->buildCustomerData($isOpenInvoice, $customer->email, $invoicingAddress->phone, $invoicingAddress->firstname, $invoicingAddress->lastname, $invoicingAddressCountryCode, $localeCode, $customer->id, $request);
+
+        if ($isOpenInvoice) {
+            $request = $this-> buildOpenInvoiceLines($paymentMethod, $cart, $request);
+        }
+
+        echo '<pre>';
+        var_dump($request);
+        echo '<pre>';
+        exit;
```
msilvagarcia: oh hai

rikterbeek: ah remind me of old days. Can't we return it in a better way ?

service/Payment.php
```diff
@@ -0,0 +1,62 @@
...
+namespace Adyen\PrestaShop\service;
+
+class Payment
+{
+    const AFTERPAY_PAYMENT_METHOD = 'afterpay';
+    const RATEPAY_PAYMENT_METHOD = 'ratepay';
+    const KLARNA_PAYMENT_METHOD = 'klarna';
+    const KLARNA_B2B_PAYMENT_METHOD = 'klarna_b2b';
+    const KLARNA_ACCOUNT_PAYMENT_METHOD = 'klarna_account';
+    const KLARNA_PAYNOW_PAYMENT_METHOD = 'klarna_paynow';
```
rikterbeek: as more and more is coming can't we regex on klarna ?

service/Payment.php
```diff
@@ -0,0 +1,62 @@
...
+ * Copyright (c) 2020 Adyen B.V.
+ * This file is open source and available under the MIT license.
+ * See the LICENSE file for more info.
+ */
+
+namespace Adyen\PrestaShop\service;
+
+class Payment
+{
+    const AFTERPAY_PAYMENT_METHOD = 'afterpay';
```
rikterbeek: afterpay has multiple products can't we regex on afterpay ? (now missing afterpay_default and afterpay_default_nl and afterpay_touch.

service/builder/Address.php
```diff
@@ -0,0 +1,172 @@
...
+        $addressType,
+        $street,
+        $houseNumberOrName,
+        $postalCode,
+        $city,
+        $stateOrProvince,
+        $country,
+        $request
+    ) {
+        if ('' !== $street) {
```
rikterbeek: can't we not better use the empty check ?

controllers/front/Payment.php
cyattilakiss: haha this is for dev only, since the logger I made is also in WIP and not even in the sprint now this is a """"nice"""" way of printing the data you need to see ;)
Please ignore it for :)

service/Payment.php
cyattilakiss: It is possible yes although learning from the Magento2 plugin experiences I would keep a list of the full payment method names instead of using regexes because there are always exceptions.

A good example: https://github.com/Adyen/adyen-magento2/blob/78c4f4af3f5b183e363214007254c075e287da37/Gateway/Request/CheckoutDataBuilder.php#L131 we check if the payment method is open invoice pm and also avoid afterpay_touch. 

Another exmaple for the confusion in Magento2: https://github.com/Adyen/adyen-magento2/pull/615

It will be harder to maintain because every time we are going to have the new open invoice payment methods into the list but with this approach we can make sure that we are in control of the payment methods to be categorised open invoice methods.

service/Payment.php
cyattilakiss: I will add them as well into the list. Do we have a precise list of payment method codes which we want to support?

service/builder/Address.php
cyattilakiss: yes that is a better idea

cyattilakiss:
This is a [WIP] PR, pushed only with a purpose of publishing it because I'm waiting for info. The ticket is moved to on hold. Of course feedback is appreciated, so meanwhile I can make the changes while waiting for the necessary input.

cyattilakiss:
I will create a meeting to review this PR.

service/builder/Address.php
```diff
@@ -0,0 +1,172 @@
...
+ */
+
+namespace Adyen\PrestaShop\service\builder;
+
+class Address extends Builder
+{
+    /**
+     * @var int
+     */
+    private static $addressTypeBilling = 1;
```
cyattilakiss:
```suggestion
    private static $addressTypeBilling = 'billingAddress';
```

service/builder/Address.php
```diff
@@ -0,0 +1,172 @@
...
+
+    /**
+     * @var int
+     */
+    private static $addressTypeDelivery = 2;
```
cyattilakiss:
```suggestion
    private static $addressTypeDelivery = 'deliveryAddress';
```

service/builder/Address.php
```diff
@@ -0,0 +1,172 @@
...
+        } else {
+            $address["country"] = self::$defaultCountry;
+        }
+
+        // Assigns the address to billing or delivery address depends on the $addressType parameter
+        if (self::$addressTypeDelivery == $addressType) {
+            $request['deliveryAddress'] = $address;
+        } elseif (self::$addressTypeBilling == $addressType) {
+            $request['billingAddress'] = $address;
+        }
```
cyattilakiss: replace with the new values

service/Gender.php
```diff
@@ -0,0 +1,55 @@
...
+     * @return mixed
+     * @throws \Exception
+     */
+    public function getAdyenGenderValueById($genderId)
+    {
+        if (isset(self::$genderMap[$genderId])) {
+            return self::$genderMap[$genderId];
+        }
+
+        throw new \Exception('Adyen gender value not found by id: ' . (int)$genderId);
```
msilvagarcia: Please improve this exception message

service/Gender.php
```diff
@@ -0,0 +1,55 @@
...
+ */
+
+namespace Adyen\PrestaShop\service;
+
+class Gender
+{
+    const MALE_ID = 1;
+    const FEMALE_ID = 2;
+    const MALE_VALUE = 'M';
+    const FEMALE_VALUE = 'F';
```
cyattilakiss: Add full values: 'MALE', 'FEMALE' and 'UNKNOWN' fall back to 'UNKNOWN'

adyen.php
```diff
@@ -1127,6 +1119,7 @@ private function getStandardPaymentMethod()
      * @param array $paymentMethod
      *
      * @return bool
+     * @deprecated don't use this to filter the supported methods anymore
      */
     private function isSimplePaymentMethod($paymentMethod)
```
cyattilakiss: we can remove it since this is a private function

helper/Data.php
```diff
@@ -307,6 +307,7 @@ public function buildControllerResponseJson($action, $details = array())
      * @param $amount
      * @param $currency
      * @return string
+     * @deprecated Use Adyen\Util\Currency sanitize() from Adyen php api library
```
cyattilakiss: Add readme that in the next major release the deprecated functions will be removed

cyattilakiss: Add "will be removed in v2.0.0"

controllers/front/Payment.php
```diff
@@ -173,21 +209,27 @@ public function postProcess()
...
+        $this->handlePaymentsResponse($response, $cart, $customer, $isAjax);
+
+        return $response;
+    }
 
+    private function handlePaymentsResponse($response, $cart, $customer, $isAjax)
+    {
+        $resultCode = $response['resultCode'];
+
+        $extra_vars = array();
```
cyattilakiss: camel case

PR #65, opened 2020-03-09
[PW-1892] Refactor using Sessions 

cyattilakiss:
Save payment response details into database instead of sessions


controllers/FrontController.php
```diff
@@ -72,4 +122,356 @@ protected function ajaxRender($value = null, $controller = null, $method = null)
...
+        }
+
+        $total = (float)$cart->getOrderTotal(true, \Cart::BOTH);
+
+        // Based on the result code start different payment flows
+        switch ($resultCode) {
+            case 'Authorised':
+                $this->module->validateOrder(
+                    $cart->id,
+                    2,
```
msilvagarcia: Apparently you need to get this number with `Configuration::get`instead

controllers/front/Result.php
```diff
@@ -27,33 +27,50 @@
...
     /**
-     * @throws \Adyen\AdyenException
+     * @throws Adapter_Exception
+     * @throws AdyenException
      */
     public function postProcess()
     {
-        if (!isset($_SESSION['paymentData'])) {
+        // Retrieve cart based on the reference parameter
+        $cart = new \Cart(\Tools::getValue('reference'));
```
msilvagarcia: This `reference` seems too generic and I fear it might be overridden

model/AdyenNotification.php
```diff
@@ -0,0 +1,216 @@
...
+
+        $originalReference = null;
+        if (!empty($notification['originalReference'])) {
+            $originalReference = trim($notification['originalReference']);
+            $sql .= ' AND `original_reference` = "' . pSQL($originalReference) . '"';
+        }
+
+        return $this->dbInstance->getValue($sql);
+    }
+}
```
msilvagarcia: I really love this entire file. Thanks for putting the effort in this kind of organization.

service/notification/NotificationProcessor.php
```diff
@@ -74,87 +81,43 @@ class NotificationProcessor
...
-            . ' SET `processing` = 0, `done` = 0'
-            . ' WHERE `done` = 0'
-            . ' AND `processing` = 1'
-            . ' AND `entity_id` = "' . (int)$id . '"';
-        return $this->dbInstance->execute($sql);
+        switch ($unprocessedNotification['eventCode']) {
+            case AdyenNotification::AUTHORISATION:
+            case AdyenNotification::OFFER_CLOSED:
+                $this->adyenPaymentResponse->deletePaymentResponseByCartId($unprocessedNotification);
+                break;
```
msilvagarcia: I would prefer to see an `if` statement instead.

adyen.php
```diff
@@ -87,10 +87,15 @@ class Adyen extends PaymentModule
...
+    /**
+     * @var Adyen\PrestaShop\model\AdyenPaymentResponse
+     */
+    private $adyenPaymentResponseModel;
+
     /**
      * Adyen constructor.
      *
-     * @throws \PrestaShop\PrestaShop\Adapter\CoreException
+     * @throws Adapter_Exception
```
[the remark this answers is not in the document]
cyattilakiss: This is the exception for 1.6 you can find it in /Adapter/Adapter_Exception.php
I usually work on 1.6 that's why the auto phpdoc changed to this - it would be a good idea to stick to one version (1.7 I would say)

adyen.php
```diff
@@ -994,8 +1043,8 @@ public function hookPaymentReturn()
     }
 
     /**
-     *
-     * @throws \PrestaShop\PrestaShop\Adapter\CoreException
+     * @return |null
+     * @throws Exception
```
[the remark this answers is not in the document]
cyattilakiss: The same as before "I usually work on 1.6 that's why the auto phpdoc changed to this"

controllers/FrontController.php
cyattilakiss: PS_OS_PAYMENT from the Configuration table you mean?
Yes that would be a great idea - easier to understand than just a 2 in the middle of the code ;)

controllers/front/Result.php
cyattilakiss: maybe merchantReference? 
I was thinking about the same with the reference but I also didn't want to include our naming

model/AdyenNotification.php
cyattilakiss: Thanks for noticing ;)

service/notification/NotificationProcessor.php
cyattilakiss: The only reason I created the switch because when we implement new and new notification types we are going to move to switch like in all the other plugins - so I just immediately moved there.

cyattilakiss:
@msilvagarcia Thanks for the feedback.
Yes this is massive, indeed :D I understand you loosing the focus. If it was not written by me probably would have happened to me as well. (or still does :) )
I hope you had a nice holiday and flight back btw ;)

controllers/FrontController.php
msilvagarcia: Not sure which value exactly, but yes :)

controllers/front/Result.php
msilvagarcia: `merchantReference` seems a bit better, although I was thinking about prefixing it with `adyen`. So `adyenMerchantReference`?

service/notification/NotificationProcessor.php
msilvagarcia: While I understand the reasoning, we could achieve a similar structure with early returns instead. But then it becomes a matter of preference I think.

My problem with long `switch` statements is that it's harder to figure out what is being compared since the variable is only at the top of the statement. To write `if` statements you need to add the variable in it, which makes it more verbose.

I'll leave the decision to you because of the deadline :)

adyen.php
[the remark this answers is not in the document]
msilvagarcia: For annotations I would stick to 1.7 as well.

service/notification/NotificationProcessor.php
cyattilakiss: Let's leave it like this now and we gonna discuss together with Angel how to refactor it in all the plugins so we can follow a standard approach :)

controllers/front/Result.php
cyattilakiss: I like the adyenMerchantReference - going to switch to that

PR #67, opened 2020-03-18
[PW-1732] Banktransfer support

cyattilakiss:
Handle Received result code
Show bank transfer details on order confirmation page

### DIFF
--- a/views/templates/front/order-confirmation.tpl
+++ b/views/templates/front/order-confirmation.tpl
@@ -23,18 +23,18 @@
 
 {include './adyencheckout.tpl'}
 
-{if $action or $additionalData}
+{if isset($action) or isset($additionalData)}
     <div class="row">
       <div class="col-md-12">
         <h3 class="card-title h3">Please use these details to finish the payment:</h3>
-        {if $action}
+        {if isset($action)}
           <div
                   data-adyen-payment-action-container
                   data-adyen-payment-action="{$action|escape:'html'}"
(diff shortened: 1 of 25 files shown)


views/templates/front/order-confirmation.tpl
```diff
@@ -23,18 +23,18 @@
 
 {include './adyencheckout.tpl'}
 
-{if $action or $additionalData}
+{if isset($action) or isset($additionalData)}
```
rikterbeek: shall we use !empty instead ?

views/templates/front/order-confirmation.tpl
```diff
@@ -23,18 +23,18 @@
...
     <div class="row">
       <div class="col-md-12">
         <h3 class="card-title h3">Please use these details to finish the payment:</h3>
-        {if $action}
+        {if isset($action)}
```
rikterbeek: shall we use !empty instead ?

views/templates/front/order-confirmation.tpl
```diff
@@ -23,18 +23,18 @@
...
           <div
                   data-adyen-payment-action-container
                   data-adyen-payment-action="{$action|escape:'html'}"
           ></div>
         {/if}
 
-        {if $additionalData}
+        {if isset($additionalData)}
```
rikterbeek: shall we use !empty instead ?

PR #73, opened 2020-04-02
[PW-2058] Fix prestashop plugin integration with multi stores setup 

cyattilakiss:
Replace path in addCSS() and registerJavascript() to support multistores
$this->_path contains the store which works only for 1 store and not for
multistore setups

### DIFF
--- a/adyenofficial.php
+++ b/adyenofficial.php
@@ -522,7 +522,7 @@ class AdyenOfficial extends PaymentModule
      */
     public function displayForm()
     {
-        $this->context->controller->addCSS($this->_path . 'views/css/adyen_admin.css', 'all');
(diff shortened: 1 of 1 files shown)


adyenofficial.php
```diff
@@ -522,7 +522,7 @@ public function displayGetStarted()
      */
     public function displayForm()
     {
-        $this->context->controller->addCSS($this->_path . 'views/css/adyen_admin.css', 'all');
+        $this->context->controller->addCSS('modules/' . $this->name . '/views/css/adyen_admin.css', 'all');
```
rikterbeek, line 525: shall we make this a constant and reuse that one on all the places ?

cyattilakiss: I would keep it this way because I need to use $this->name in the value so I could create a static property maybe.

PR #72, opened 2020-04-01
[PW 2234]: Configuration panel header

acampos1916:
Here we're adding a header with useful links to the module's configuration panel

PR #76, opened 2020-04-29
Release 2.1.1

cyattilakiss:
#72 [PW 2234]: Configuration panel header 
#73 [PW-2058] Fix plugin integration with multi stores setup
#74 [PW-1890] Update ReadMe 
#75 [PW-2343] Meet prestashop validation requirements 
#80 [PW-2235]: Adding translations to smarty templates 
#81 [PW-2377] Fix cron job controller name and url


adyenofficial.php
```diff
@@ -92,6 +92,30 @@ class AdyenOfficial extends PaymentModule
...
+            "label" => "Support",
+            "url" => "https://support.adyen.com/hc/en-us/requests/new?ticket_form_id=78764"
+        ),
+        array(
+            "label" => "GitHub",
+            "url" => "https://github.com/Adyen/adyen-prestashop/releases"
+        ),
+        array(
+            "label" => "PrestaShop Marketplace",
+            "url" => "https://addons.prestashop.com"
```
acampos1916, line 115: Do we really have a marketplace link for Adyen? I cannot find it on addons.prestashop.com so we can update it here

cyattilakiss: In order to find your module, please make sure that your location (at the bottom of the page) is set on the countries of France or Spain.
I updated the link, thanks for noticing!

PR #84, opened 2020-05-15
[PW-2241] Update processing Authorisation and Offer_Closed notifications

cyattilakiss:
- Take success field into consideration when processing the
 notifications
 - Add getProcessedNotificationsByMerchantReference() to retrieve a list
 of already processed notifications for an order by merchant_reference
 - Add hasProcessedAuthorisationSuccessNotification() to check if an
 order has already had an Authorised notification with success = true
 value processed

### DIFF
--- a/model/AdyenNotification.php
+++ b/model/AdyenNotification.php
@@ -213,4 +213,17 @@ class AdyenNotification extends AbstractModel
 
         return $this->dbInstance->getValue($sql);
     }
+
+    /**
+     * @param $merchantReference
+     * @return mixed
+     */
+    public function getProcessedNotificationsByMerchantReference($merchantReference)
+    {
+        $sql = 'SELECT * FROM ' . _DB_PREFIX_ . self::$tableName
+            . ' WHERE `merchant_reference` = "' . pSQL($merchantReference) . '"'
+            . ' AND `done` = 1';
+
+        return $this->dbInstance->executeS($sql);
+    }
 }
--- a/service/notification/NotificationProcessor.php
+++ b/service/notification/NotificationProcessor.php
@@ -129,11 +129,38 @@ class NotificationProcessor
 
         switch ($unprocessedNotification['event_code']) {
             case AdyenNotification::AUTHORISATION:
-                $order->setCurrentState(\Configuration::get('PS_OS_PAYMENT'));
-                $this->orderService->addPaymentDataToOrderFromResponse($order, $unprocessedNotification);
+                if ('true' === $unprocessedNotification['success']) {
+                    // AUTHORISATION success = true moves order to paid if order status is not PS_OS_PAYMENT
+                    if ($order->getCurrentState() !== \Configuration::get('PS_OS_PAYMENT')) {
+                        $order->setCurrentState(\Configuration::get('PS_OS_PAYMENT'));
+                    }
+
+                    $this->orderService->addPaymentDataToOrderFromResponse($order, $unprocessedNotification);
+                } else {
+                    if ($order->getCurrentState() !== \Configuration::get('PS_OS_CANCELED')) {
+                        // No previous authorisation notification was processed before
+                        if (!$this->hasProcessedAuthorisationSuccessNotification(
+                            $unprocessedNotification['merchant_reference']
+                        )) {
+                            // AUTHORISATION success = false moves order to canceled if not canceled already
+                            $order->setCurrentState(\Configuration::get('PS_OS_CANCELED'));
+                        } else {
+                            $this->logger->addAdyenNotification('Notification with entity_id (' . $unprocessedNotification['entity_id'] . ') was ignored during processing the notifications because another Authorisation success = true notification has already been processed for the same order.');
+                        }
+                    }
+                }
+
                 $this->adyenPaymentResponse->deletePaymentResponseByCartId($unprocessedNotification);
                 break;
             case AdyenNotification::OFFER_CLOSED:
+                if ('true' === $unprocessedNotification['success']) {
+                    // OFFER_CLOSED success = true moves order to canceled if order status is
+                    // ADYEN_OS_WAITING_FOR_PAYMENT
+                    if ($order->getCurrentState() === \Configuration::get('ADYEN_OS_WAITING_FOR_PAYMENT')) {
+                        $order->setCurrentState(\Configuration::get('PS_OS_CANCELED'));
+                    }
+                }
+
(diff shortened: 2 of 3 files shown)


service/notification/NotificationProcessor.php
```diff
@@ -129,11 +129,41 @@ public function processNotification($unprocessedNotification)
 
         switch ($unprocessedNotification['event_code']) {
             case AdyenNotification::AUTHORISATION:
-                $order->setCurrentState(\Configuration::get('PS_OS_PAYMENT'));
-                $this->orderService->addPaymentDataToOrderFromResponse($order, $unprocessedNotification);
+                if ('true' === $unprocessedNotification['success']) {
```
rikterbeek, line 132: looks a bit weird should we not move this to unprocessedNotification['success'] == true or does boolean checking not working ?

service/notification/NotificationProcessor.php
```diff
@@ -129,11 +129,41 @@ public function processNotification($unprocessedNotification)
...
+                                'notifications because another Authorisation success = true notification has already ' .
+                                'been processed for the same order.');
+                        }
+                    }
+                }
+
                 $this->adyenPaymentResponse->deletePaymentResponseByCartId($unprocessedNotification);
                 break;
             case AdyenNotification::OFFER_CLOSED:
+                if ('true' === $unprocessedNotification['success']) {
```
rikterbeek, line 159: same here.

service/notification/NotificationProcessor.php
```diff
@@ -233,4 +263,34 @@ public function addMessage($notification)
...
+
+        if (empty($processedNotifications)) {
+            return false;
+        }
+
+        $hasProcessedAuthorisedNotification = false;
+
+        foreach ($processedNotifications as $processedNotification) {
+            if (AdyenNotification::AUTHORISATION === $processedNotification['event_code'] &&
+                'true' === $processedNotification['success']
```
rikterbeek, line 288: same here.

service/notification/NotificationProcessor.php
rikterbeek: should we not store it as well as a boolean value ?

service/notification/NotificationProcessor.php
```diff
@@ -233,4 +263,34 @@ public function addMessage($notification)
...
+            ) {
+                $hasProcessedAuthorisedNotification = true;
+            }
+        }
```
rikterbeek, line 292: we can add this as a method and when one item is found return it so we don't have to loop to all of the next notifications.

service/notification/NotificationProcessor.php
```diff
@@ -129,11 +129,41 @@ public function processNotification($unprocessedNotification)
...
+                    }
+
+                    $this->orderService->addPaymentDataToOrderFromResponse($order, $unprocessedNotification);
+                } else {
+                    if ($order->getCurrentState() !== \Configuration::get('PS_OS_CANCELED')) {
+                        // No previous authorisation notification was processed before
+                        if (!$this->hasProcessedAuthorisationSuccessNotification(
+                            $unprocessedNotification['merchant_reference']
+                        )) {
+                            // AUTHORISATION success = false moves order to canceled if not canceled already
```
rikterbeek, line 145: can we make this an early opt out so we reduct the number of if else statmeents. So something like:
if (!$this->hasProcessedAuthorisationSuccessNotification(
                            $unprocessedNotification['merchant_reference']
return;
                        ))

model/AdyenNotification.php
```diff
@@ -213,4 +213,17 @@ public function isDuplicate($notification)
...
+     * @return mixed
+     */
+    public function getProcessedNotificationsByMerchantReference($merchantReference)
+    {
+        $sql = 'SELECT * FROM ' . _DB_PREFIX_ . self::$tableName
+            . ' WHERE `merchant_reference` = "' . pSQL($merchantReference) . '"'
+            . ' AND `done` = 1';
+
+        return $this->dbInstance->executeS($sql);
+    }
```
rikterbeek, line 228: We can make it a static method easier to call without the need to declare the model.

service/notification/NotificationProcessor.php
cyattilakiss: The database table structure indicates the value is a string and we would like to keep this check simple that's why we don't cast it to a boolean and just deal with it as a string. We could store it of course as a boolean in the future.

model/AdyenNotification.php
cyattilakiss: It is easier to mock the function if it is not static and when we created the notification model we decided to not create static functions even though almost all the functions in that class could be static.

msilvagarcia: Static methods are harder to test in PHP. Also, PrestaShop offers a Dependency Injection suite so it's not necessary to declare the model manually.

service/notification/NotificationProcessor.php
cyattilakiss: In this exact case I believe an if else makes this scenario easier to understand. Usually I agree that the early returns are better and easier to understand for example when an array is empty then return immediately because there is no use of continue the process (example from this PR: https://github.com/Adyen/adyen-prestashop/pull/84/files/03ac798e276ede922c236a198c6d491bf69a21c4#diff-b28f8e395dca11241416e156fff07337R280-R282)

In this one we have 2 cases which depends on the function: hasProcessedAuthorisationSuccessNotification()

model/AdyenNotification.php
rikterbeek: Ok, clear.

service/notification/NotificationProcessor.php
rikterbeek: ok let's keep it is like this for now.

service/notification/NotificationProcessor.php
rikterbeek: It is now really hard to read the code. Can't we include it then in the general if else statements above and add this into the if criteria ?

rikterbeek:
@cyattilakiss check codesniffer: FILE: service/notification/NotificationProcessor.php
----------------------------------------------------------------------
FOUND 1 ERROR AFFECTING 1 LINE
----------------------------------------------------------------------
 142 | ERROR | [x] Blank line found at start of control structure
----------------------------------------------------------------------
PHPCBF CAN FIX THE 1 MARKED SNIFF VIOLATIONS AUTOMATICALLY

PR #93, opened 2020-07-15
[PW-2492] Fix refund calls in partial refunds

msilvagarcia:
## Summary

Partial refunds were not working entirely in the plugin due to mismatching order ids and invalid SQL query.

## Tested scenarios

Executing a partial refund with Credit Card and Klarna sends the refund request to Adyen servers.

**Fixed issue**:

### DIFF
--- a/infra/NotificationRetriever.php
+++ b/infra/NotificationRetriever.php
@@ -50,7 +50,7 @@ class NotificationRetriever
      */
     public function getPSPReferenceByOrderId($orderId)
     {
-        $orderId = intval($orderId);
+        $orderId = (int)$orderId;
         $results = $this->db->executeS(
             sprintf(
                 <<<SQL
--- a/infra/NotificationRetriever.php
+++ b/infra/NotificationRetriever.php
@@ -50,13 +50,14 @@ class NotificationRetriever
      */
     public function getPSPReferenceByOrderId($orderId)
     {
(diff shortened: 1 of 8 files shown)


infra/NotificationRetriever.php
```diff
@@ -50,13 +50,14 @@ public function __construct()
...
             sprintf(
                 <<<SQL
-select a.pspreference as pspReference
-from %sadyen_notification a
-inner join %sorders o on a.merchant_reference = o.id_cart
-where o.id_order = intval($orderId)
+select op.transaction_id as pspReference
+from %sorder_payment op
+inner join %sorders o on op.order_reference = o.reference
+where o.id_order = $orderId
```
cyattilakiss, line 60: could you please run the Presta validator with this change? I think last time it was throwing a security error if the variable was like this. I don't know if it is smart enough to connect it with line 53.

cyattilakiss:
Could you please create a ticket to refactor and move this function from the 'infra/NotificationRetriever.php' and rename it since it's no longer retrieving any notification only the psp reference for an order?

msilvagarcia:
Partial refunds were not working entirely in the plugin due to mismatching order ids and an invalid SQL query.

This commit also makes sure that the PSP Reference is updated when processing a notification if the payment object doesn't have it stored in `transaction_id`.

PR #96, opened 2020-07-17
[PW-1897] Help understand standard refunds and merchandise returns

msilvagarcia:
## Summary

This helps us understand and debug issues with standard refunds by:
1. Adding documentation on how to enable it
1. Refactoring the refund process
1. Adding logging for when a refund fails due to missing credit slip

## Tested scenarios

Refund orders from credit card payments

**Fixed issue**:


service/RefundService.php
```diff
@@ -0,0 +1,97 @@
...
+        $orderSlip = $order->getOrderSlipsCollection()
+                           ->orderBy('date_upd', 'desc')
+                           ->getFirst();
+        $currency = Currency::getCurrency($order->id_currency);
+        if ($orderSlip) {
+            return $refundService->request($orderSlip, $currency['iso_code']);
+        } else {
+            $this->logger->error('Refund occurred without a credit slip.');
+            return false;
+        }
```
cyattilakiss, line 95:
```suggestion
        }
        
        return false;
```

Issue #106, opened 2020-07-31
[PW-2817] 5 times the same order in the same second 


YanP-Adyen:
Hi team,

**Describe the bug**
Shopper pays at 18h10 : 1845959526403196
3 minutes later, there is a duplication of the same order for 5 transactions, all withing the same second : 
1345959528082613
1645959528081240
1345959528082647
1645959528081281
4745959528083470

As per the log I don't see anything wrong. First I thought about the shopper that 'spam clicked' the pay button but the fact that the 5 duplicates transactions comes 3 minutes after the first one invalidate this theory.
See attached : 
[adyen_notification.log](https://github.com/Adyen/adyen-prestashop/files/5007795/adyen_notification.log)
[error.log](https://github.com/Adyen/adyen-prestashop/files/5007797/error.log)
[info.log](https://github.com/Adyen/adyen-prestashop/files/5007798/info.log)

The merchant, Degriffstock, is using plugin v2.1.5 with prestashop v1.6.1.18.

Thanks.

YanP-Adyen:
The issue repeated yesterday, 3 times : 
4735963722527763
1635963722523058
(merchant ref 570698)

4555963780627221
4755963780635917
(merchant ref 570735)

1855963802601082
4755963802619154
(merchant ref 570731)

I tried clicking multiple times on "Pay button" on the pre-prod website, but it always increment the merchant reference which is not the case here.

AlexandrosMor:
Hello @YanP-Adyen,

Thank you for posting this issue, We are going to look at it and come back to you as soon as possible.

kind regards,
Alexandros

AlexandrosMor:
hello @YanP-Adyen ,

Indeed in the logs there are some transactions in the same time. I beleive it is link to #100. 

regards,
Alexandros

YanP-Adyen:
Hi @AlexandrosMor 

Thanks for your feedback.
I don't think so, I tried to do this : clicking multiple times on the Pay Button.
Indeed I find multiple transactions, but with an incremented merchant reference each time.

Here the merchant reference is always the same.

Regards,
Yan

cyattilakiss:
Hi @YanP-Adyen 

I was finally able to reproduce this issue on 1.6 and it this issue and I believe it also fixed in the #111 since after that fix I am not able to do so anymore.

After the release please let us know if it indeed works for the merchant as well.

Best,
Attila
Adyen

cyattilakiss:
[...]

PR #111, opened 2020-08-26
[PW-2779] Adding disabling button after click on Pay

### LINKED ISSUE #100: [PW-2779] Adding a spinner/disabling button after click on Pay
**Is your feature request related to a problem? Please describe.**
If there's some network latency the user could spam Pay button, this will result in multiple transactions being made

**Describe the solution you'd like**
Adding a spinner and disabling the button after the first click to avoid such issues.

**Describe alternatives you've considered**
If a spinner is not possible at least disabling the button would make it.

Merchant requirement : DegriffStock

cyattilakiss:
## Summary
If there's some network latency the user could spam Pay button in prestashop version 1.6, this will result in multiple transactions being made. Adding a spinner and disabling the button after the first click to avoid such issues.

## Tested scenarios
1.6 and 1.7
Valid, invalid cards, non3DS, 3DS2
Valid, invalid stored payment methods
Local payment methods 

**Fixed issue**:  
Fixes #100 #106

### DIFF
--- a/views/js/payment-components/card-payment-method.js
+++ b/views/js/payment-components/card-payment-method.js
@@ -28,8 +28,10 @@ jQuery(function ($) {
 
     var allValidCard;
     var placeOrderAllowed;
+    var placeOrderInProgress = false;
     var popupModal;
     var data;
+    var cardComponent;
 
     /* Create adyen checkout with default settings */
     placeOrderAllowed = false;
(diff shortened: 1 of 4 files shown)


views/js/payment-components/card-payment-method.js
```diff
@@ -253,4 +268,22 @@ jQuery(function ($) {
...
+        paymentForm.find('button[type="submit"]').prop('disabled', true);
+        paymentForm.find('button[type="submit"] i').toggleClass('icon-spinner icon-chevron-right right');
+    }
+
+    function placingOrderEnds()
+    {
+        placeOrderInProgress = false;
+        paymentForm.find('button[type="submit"]').prop('disabled', false);
+        paymentForm.find('button[type="submit"] i').toggleClass('icon-spinner icon-chevron-right right');
+    }
```
acampos1916, line 288: No way to have these functions in a common place?

cyattilakiss: Couldn't figure it out how to require a PrestaShop registered JS loaded for sure before the function from that one is called that's why the duplications.

acampos1916 (approved):

msilvagarcia (approved):

PR #112, opened 2020-08-26
[PW-2926] Plugin change 2 times the command status

### LINKED ISSUE #110: [PW-2926] Plugin change 2 times the command status
**Describe the bug**
For unknown reasons, sometimes the shopper receives 2 email confirming its order.
With the merchant we saw that prestashop/the plugin was updating 2 times the order status to "payment accepted" within a 2/3 minutes gap.
This not very urgent nor blocking anything, but it leads the shopper thinking that he will pay 2 times.

Untitled

(No need to speak french ;))
at 16h34 the order is "payment accepted"
at 16h36 we receive the notification and the payment is set to accepted again.

The idea would be to understand why there is 2 status update.

The crontable is configued as follow

` * * * * * /usr/bin/curl -k "https://preprodurl/index.php?fc=module&controller=AdminAdyenOfficialPrestashopCron&token=..." `
` * * * * * /usr/bin/curl -k "https://liveurl/index.php?fc=module&controller=AdminAdyenOfficialPrestashopCron&token=..." `

cyattilakiss:
## Summary
Only move orders to paid during processing notifications when the order has specific states 
An Authorisation success = true notification should only move an order state to paid when it's canceled, error or waiting for payment
In case the order is already paid or moved to another state, the notification effect will be ignored

With this approach multiple state switching and email sending can be avoided

**Fixed issue**:  
Fixes #110

### DIFF
--- a/service/notification/NotificationProcessor.php
+++ b/service/notification/NotificationProcessor.php
@@ -332,7 +332,8 @@ class NotificationProcessor
      * @param string $currentOrderStatus
      * @return bool
      */
-    private function isCurrentOrderStatusANonFinalStatus($currentOrderStatus) {
+    private function isCurrentOrderStatusANonFinalStatus($currentOrderStatus)
+    {
         foreach (self::$nonFinalOrderStatuses as $nonFinalOrderStatus) {
             if ($currentOrderStatus === \Configuration::get($nonFinalOrderStatus)) {
                 return true;
--- a/service/notification/NotificationProcessor.php
+++ b/service/notification/NotificationProcessor.php
@@ -39,6 +39,19 @@ use Adyen\PrestaShop\service\Order as OrderService;
 class NotificationProcessor
 {
     /**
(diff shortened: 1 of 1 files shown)


service/notification/NotificationProcessor.php
```diff
@@ -312,4 +325,20 @@ private function hasProcessedAuthorisationSuccessNotification($merchantReference
...
         return false;
     }
+
+    /**
+     * Checks if the current order status is in the self::$nonFinalOrderStatuses list
+     *
+     * @param string $currentOrderStatus
+     * @return bool
+     */
+    private function isCurrentOrderStatusANonFinalStatus($currentOrderStatus) {
```
acampos1916, line 335:
```suggestion
    private function isCurrentOrderStatusANonFinalStatus($currentOrderStatus)
    {
```

acampos1916 (dismissed):

acampos1916 (approved):

msilvagarcia (approved):

PR #128, opened 2020-10-14
[PW-3181] 3DS1 same site cookie fix

cyattilakiss:
## Summary
**Draft until the issue can be reproduced on test environment and can be tested properly.**

A new redirect page is created which does not retrieve or regenerate the session cookie but only receives the POST requests and submits it via POST to the original Result url. GET requests are being redirected without any change to the Result endpoint.

## Tested scenarios

3DS1 enabled test card
iDeal redirect payment method

**Fixed issue**:


cyattilakiss:
Closing it since it seems like we don't have this problem!

Issue #132, opened 2020-10-25
[PW-3329] Call to undefined method ModuleFrontController::ajaxRender


mehdi-ghezal:
**Describe the bug**
After installing the module and following the guideline about configuration for the test environment, when I pay with a test card, the AJAX request sent by the module lead to a PHP  Fatal Error.

**To Reproduce**

1. Create a new account on Adyen
1. Install the Prestashop module directly from Prestashop Backoffice
1. Follow the setup instruction: https://docs.adyen.com/plugins/prestashop
1. Proceed to checkout and "Pay by Carte bancaire"
1. Use the test card from: https://docs.adyen.com/development-resources/test-cards/test-card-numbers ; I use this one: 4035 5010 0000 0008 - Visa Debit / Cartes Bancaires - FR - 03/2030 - 737

**Expected behavior**
Payement successful or at least a feedback error ; obviously not a Fatal Error.

**Logs**

```
PHP Fatal error:  Uncaught Error: Call to undefined method ModuleFrontController::ajaxRender() in ...../modules/adyenofficial/controllers/FrontController.php:155
Stack trace:
    #0 ...../modules/adyenofficial/controllers/front/Payment.php(191): Adyen\PrestaShop\controllers\FrontController->ajaxRender('{"action":"erro...')
    #1 ...../classes/controller/Controller.php(242): AdyenOfficialPaymentModuleFrontController->postProcess()
    #2 ...../classes/Dispatcher.php(428): ControllerCore->run()
    #3 ...../index.php(28): DispatcherCore->dispatch()
    #4 {main}
        thrown in ...../modules/adyenofficial/controllers/FrontController.php on line 155
``` 

**Desktop (please complete the following information):**
Not applicable/relevant

**Smartphone (please complete the following information):**
Not applicable/relevant

**Additional context**
Linux Debian 9.13 with PHP 7.0 (PHP FPM)
Prestashop 1.7.3

cyattilakiss:
Hi @mehdi-ghezal ,

Thank you for raising this issue.
I could verify that it's indeed an issue on the PrestaShop version 1.7.3 and was fixed in the version 1.7.5.
I will create and assign an internal ticket to fix this incompatibility issue in those versions asap.

Best,
Attila
Adyen

cyattilakiss:
Hi @mehdi-ghezal ,

The issue has been fixed in the release 3.2.0
I will close this issue for now but please let us know if you still face this issue or have any further questions.

Best,
[...]

PR #133, opened 2020-10-26
[PW-3329] Call to undefined method ModuleFrontController::ajaxRender

cyattilakiss:
## Summary
Use ajaxRender when available otherwise fall back to ajaxDie
ajaxRender has been introduced in prestashop version 1.7.5 and not in
1.7, before 1.7.5 only ajaxDie is available

## Tested scenarios
card payments with 1.7 and 1.6

**Fixed issue**:  
#132 [PW-3329] Call to undefined method ModuleFrontController::ajaxRender

### DIFF
--- a/controllers/FrontController.php
+++ b/controllers/FrontController.php
@@ -148,12 +148,12 @@ abstract class FrontController extends \ModuleFrontController
     protected function ajaxRender($value = null, $controller = null, $method = null)
     {
         header('content-type: application/json; charset=utf-8');
-        if ($this->versionChecker->isPrestaShop16()) {
-            $this->ajax = true;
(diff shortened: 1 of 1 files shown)


acampos1916:
Looks like `method_exists()` doesn't work the same in <=5.4?

cyattilakiss:
> Looks like `method_exists()` doesn't work the same in <=5.4?

No it's the ::class which doesn't exist under 5.5 :( I will have to change it.
"The special ::class constant is available as of PHP 5.5.0, and allows for fully qualified class name resolution at compile time, this is useful for namespaced classes:"

acampos1916 (approved):

msilvagarcia (approved):

PR #134, opened 2020-11-10
Add Giropay support

cyattilakiss:
## Summary
Support Giropay via simple redirect instead of component

## Tested scenarios

1.6 and 1.7

### DIFF
--- a/views/js/checkout-component-renderer.js
+++ b/views/js/checkout-component-renderer.js
@@ -223,13 +223,13 @@ function renderPaymentMethods() {
             }
 
             if (!placeOrderAllowed) {
-                if (!!component) {
+                if (!!component && 'showValidation' in component) {
                     component.showValidation();
                 }
 
                 return;
             }
-            
+
             if (isPlaceOrderInProgress()) {
                 return false;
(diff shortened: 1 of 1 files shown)


Aleffio (approved):

views/js/checkout-component-renderer.js
```diff
@@ -223,10 +223,13 @@ function renderPaymentMethods() {
             }
 
             if (!placeOrderAllowed) {
-                component.showValidation();
+                if (!!component) {
```
msilvagarcia, line 226:
```suggestion
                if (component && component.showValidation) {
```

msilvagarcia (changes requested):
Does my suggestion for line 226 offer the same validation?

views/js/checkout-component-renderer.js
cyattilakiss: The show validation is a function of every component, is it needed to check in the if `component.showValidation` as well?

Aleffio (approved):

msilvagarcia (approved):

PR #139, opened 2020-11-20
Move CI to GitHub actions

msilvagarcia:
## Summary

As previously decided.

## Tested scenarios

**Fixed issue**:


.github/workflows/syntax-check.yml
```diff
@@ -0,0 +1,24 @@
...
+on: [push, pull_request]
+
+jobs:
+  build:
+    runs-on: ${{ matrix.operating-system }}
+
+    strategy:
+      matrix:
+        operating-system: [ubuntu-16.04, ubuntu-latest]
+        php-version: [5.6, 7.0, 7.1, 7.2, 7.3, 7.4, 8.0]
```
cyattilakiss, line 12: is it possible to still keep the 5.3, 5.4, 5.5 in the matrix until we are able to deprecate these versions?

msilvagarcia: Done!

Aleffio (approved):

cyattilakiss (approved):

PR #144, opened 2020-12-15
Fix `shopperReference` length

msilvagarcia:
## Summary

If `shopperReference` is shorter than 3 characters, Adyen will not store payment methods even when asked to do so. This commit fixes this issue by prepending characters to the field without breaking the feature for shoppers who already have cards stored.

## Tested scenarios

* Payment with credit card while asking to save the payment method.
* See stored payment method in payments list.

**Fixed issue**:

### DIFF
--- a/helper/Data.php
+++ b/helper/Data.php
@@ -38,9 +38,6 @@ use Country;
 use Currency;
 use Exception;
 use FileLogger;
-use Guest;
-use PrestaShopDatabaseException;
-use PrestaShopException;
 use Tools;
 
 class Data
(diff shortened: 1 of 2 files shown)


acampos1916 (approved):

cyattilakiss (approved):

msilvagarcia:
If `shopperReference` is shorter than 3 characters, Adyen will not store payment methods even when asked to do so. This commit fixes this issue by prepending characters to the field without breaking the feature for shoppers who already have cards stored.

PR #149, opened 2021-01-22
[PW-3511] Disable apple pay button when order cannot be placed

Morerice:
## Summary

- Implement onClick callback and reject in case the terms and conditions are not checked
- Hide apple pay complete payment method in case apple pay isAvailable is false (also google pay and other methods with isAvailable)
- Fix prestashop css issue which was overriding the apple pay button

## Tested scenarios

Button tested on both GooglePay and ApplePay on 1.6 and 1.7 (ApplePay payment flow could not be tested due to simulator bug on Catalina).


views/css/adyen.css
```diff
@@ -92,3 +92,9 @@ div.adyen-payment-method-label {
 div.adyen-additional-data-line {
     padding-left: 15px;
 }
+
+@supports (-webkit-appearance:-apple-pay-button) {
+    .adyen-checkout__applepay__button._1dIj5OBNHw-_NvtfqKwJuZ {
```
cyattilakiss, line 97: do not use this id `_1dIj5OBNHw-_NvtfqKwJuZ` it might change in the next bundle release
Also please add a comment above this style why it was needed

views/js/checkout-component-renderer.js
```diff
@@ -504,6 +491,17 @@ jQuery(document).ready(function() {
             }
         }
 
+        function handleOnClick(resolve, reject) {
+            const paymentMethodContainer = this.paymentForm.closest('.adyen-payment');
+            // Show message if button is disabled else if not in progress, hide and resolve
+            if (prestaShopPlaceOrderButton.prop('disabled') && !isPlaceOrderInProgress()) {
+                showRequiredConditionsInfoMessage(paymentMethodContainer);
```
cyattilakiss, line 498: why not using the `reject()`?

views/css/adyen.css
Morerice: CSS selector updated

views/js/checkout-component-renderer.js
Morerice: Reject function added

views/js/checkout-component-renderer.js
```diff
@@ -504,6 +491,18 @@ jQuery(document).ready(function() {
...
+                reject(new Error('Terms of service not agreed'));
```
cyattilakiss, line 499: is this shown on the frontend or only in the console?

Morerice: Just in the console

cyattilakiss (approved):

msilvagarcia (approved):

Issue #130, opened 2020-10-21
[PW-3316] Incoherence between order status from notifications system and payment response handling


SammyCOLIN:
**Description**
When an order is created with a Adyen payment, the order status is "payement accepted".
If an error has occured during the payment, for example "authentication failed (3D-secure)", but the payment is OK after the second attempt, we receive 2 notifications : one AUTHORISATION with **success FALSE** and another AUTHORISATION with **success TRUE**. So the Notification system (NotificationProcessor) move the order from "Payment accepted" to "Canceled" (first notification) and then from "Canceled" to "Payment accepted" (second notification). Consequently the customer receive one mail with "order OK" (when the order is created) and another with "order canceled" (when the first success false notification is processed).

**To Reproduce**
1. Create an order
2. Select Adyen payment method
3. 1st attempt : make an error during the 3D secure check (example : wrong code)
4. 2nd attempt : make a success payment
5. Redirection on the seller site : order OK (payment OK)
6. We receive a mail : order ok
7. 1 minute after we receive another mail : order canceled (first notif processed, cron job configured to each minute).

**Expected behavior**
Why the order is not set to status "Waiting payment" after the payment in order to let the notification system to set canceled/payment ok ?

**Screenshots**
Prestashop back-office
![2020-10-20_15h26_21](https://user-images.githubusercontent.com/73232938/96721794-ec053900-13ac-11eb-8a7e-08fb5b1e58df.png)

Notifications received (database)
![2020-10-20_15h29_17](https://user-images.githubusercontent.com/73232938/96721818-f3c4dd80-13ac-11eb-8c39-39983321e5ed.png)

**Context**

- Adyen module 2.1.7

- Prestashop 1.6.1.20

cyattilakiss:
Hi @SammyCOLIN ,

Thank you for raising the issue. 
Indeed this happens sometimes that you receive a success = false and a success = true AUTHORISATION notification after each other. This happens when the issuer detects a 3DS1 failure first - like closing and reopening the mobile app. Unfortunately the notifications we cannot fix because it's something that the issuer sends to Adyen and Adyen forwards it to the plugin. What we can do though is not to change the order status when the authorisation success = false notification arrives because until then we already know that the payment is authorised therefore the notification can be ignored an the customer won't get an update about that this way.

We are going to investigate the above fix and get back to you soon.

Best,
Attila
Adyen

cyattilakiss:
Hi @SammyCOLIN ,

[...]

PR #152, opened 2021-01-28
[PW-3316] Incoherence between order status from notifications system and payment response handling

Morerice:
## Summary
- Always prioritize the data received from the payment details call, over the data received in the notification call.
- Delay the processing of an `authorization w/success = false` notification by an hour 

## Tested scenarios
1. Successful authorization notification processed after order is already set to successful. No changes.
2. Not successful authorization notification processed after order is already set to successful. Order status **not** updated
3. Not successful authorization notification processed after order is already set to cancelled. No changes.

**Fixed issue**:  
#130

### DIFF
--- a/service/notification/NotificationProcessor.php
+++ b/service/notification/NotificationProcessor.php
@@ -181,10 +181,8 @@ class NotificationProcessor
                 } else { // Notification success is 'false'
                     // Order state is not canceled yet
                     if ($order->getCurrentState() !== \Configuration::get('PS_OS_CANCELED')) {
-                        // No previous authorisation notification was processed before
-                        if (!$this->hasProcessedAuthorisationSuccessNotification(
-                            $unprocessedNotification['merchant_reference']
-                        )) {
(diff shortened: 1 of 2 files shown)


acampos1916 (approved):

service/notification/NotificationProcessor.php
```diff
@@ -206,7 +203,7 @@ public function processNotification($unprocessedNotification)
                 
                 break;
             case AdyenNotification::OFFER_CLOSED:
-                // Notification success is 'true'
+                // Notification success is 'true' AND current status is a non-final one
```
cyattilakiss, line 206:
```suggestion
                // Notification success is 'true' AND current status is ADYEN_OS_WAITING_FOR_PAYMENT
```

Offer_closed should not cancel even other statuses only the waiting for payment one

cyattilakiss (dismissed):

cyattilakiss (approved):

acampos1916 (approved):

PR #153, opened 2021-02-03
[PW-2767] - New paypal integration 

Morerice:
## Summary

New paypal integration on prestashop. The main difference to other payment methods was that we were required to use the `handleAction` and not the `createFromAction`.

The other main change was to implement the `onCancel` function which sends a request to the payment details endpoint with an additional parameter. Because of this parameter, the back end will set the order to cancelled and return an appropriate message.

## Tested scenarios
Happy flow on 1.6 and 1.7
Unhappy flow on 1.6 and 1.7

### DIFF
--- a/controllers/FrontController.php
+++ b/controllers/FrontController.php
@@ -227,9 +227,10 @@ abstract class FrontController extends \ModuleFrontController
                 // PaymentResponse can be deleted
                 $this->adyenPaymentResponseModel->deletePaymentResponseByCartId($cart->id);
 
-                // In case of refused payment there is no order created and the cart needs to be cloned and reinitiated
+                // In case of refused/cancelled payment there is no order created and the cart needs to be cloned and
+            // reinitiated
                 $this->cartService->cloneCurrentCart($this->context, $cart);
-                $this->logger->error("The payment was refused, id:  " . $cart->id);
+                $this->logger->error('The payment was ' . strtolower($resultCode) . ', id:  ' . $cart->id);
 
                 if ($cancelled) {
                     $message = $this->l('The payment was cancelled by the customer');
--- a/controllers/FrontController.php
+++ b/controllers/FrontController.php
@@ -229,7 +229,7 @@ abstract class FrontController extends \ModuleFrontController
                 // In case of refused/cancelled payment there is no order created and the cart needs to be cloned and
                 // reinitiated
                 $this->cartService->cloneCurrentCart($this->context, $cart);
-                $this->logger->error('The payment was ' . strtolower($resultCode) . ', id:  ' . $cart->id);
+                $this->logger->error('The payment was ' . strtolower($resultCode) . ', with cart id:  ' . $cart->id);
 
                 if ($resultCode === 'Cancelled') {
                     $message = $this->l('The payment was cancelled by the customer');
--- a/controllers/FrontController.php
+++ b/controllers/FrontController.php
@@ -337,11 +337,12 @@ abstract class FrontController extends \ModuleFrontController
                 );
 
                 if ($isAjax) {
+                    $message = $this->l('There was an error with the payment method, please choose another one');
                     $this->ajaxRender(
                         $this->helperData->buildControllerResponseJson(
                             'error',
                             array(
-                                'message' => $this->l('There was an error with the payment method, please choose another one'),
+                                'message' => $message,
                             )
                         )
                     );
--- a/controllers/FrontController.php
+++ b/controllers/FrontController.php
@@ -170,9 +170,10 @@ abstract class FrontController extends \ModuleFrontController
      * @param $cart
      * @param $customer
      * @param $isAjax
+     * @param $cancelled
      * @throws AdyenException
      */
-    protected function handlePaymentsResponse($response, $cart, $customer, $isAjax)
+    protected function handlePaymentsResponse($response, $cart, $customer, $isAjax, $cancelled = false)
     {
         $resultCode = $response['resultCode'];
 
@@ -222,6 +223,7 @@ abstract class FrontController extends \ModuleFrontController
 
                 break;
             case 'Refused':
+            case 'Cancelled':
                 // PaymentResponse can be deleted
                 $this->adyenPaymentResponseModel->deletePaymentResponseByCartId($cart->id);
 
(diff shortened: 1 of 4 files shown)


views/js/checkout-component-renderer.js
```diff
@@ -492,17 +497,49 @@ jQuery(document).ready(function() {
...
+            const paymentMethodType = paymentMethodContainer.data('local-payment-method');
             // Show message if button is disabled else if not in progress, hide and resolve
             if (prestaShopPlaceOrderButton.prop('disabled') && !isPlaceOrderInProgress()) {
                 showRequiredConditionsInfoMessage(paymentMethodContainer);
-                reject(new Error('Terms of service not agreed'));
+                if (paymentMethodType === 'paypal') {
+                    return false;
+                } else {
+                    reject(new Error('Terms of service not agreed'));
+                }
```
acampos1916, line 509: If I understand correctly Paypal won't support T&Cs? If that's the case let's make a mention about it in the PR description so that it could be added to the docs upon release @Morerice

acampos1916 (approved):

views/js/checkout-component-renderer.js
cyattilakiss: Actually there is a bug in the current version fo the component and this is the workaround to support the onClick callback. But it does the same like the other components we just need to return false instead of reject()

views/js/checkout-component-renderer.js
```diff
@@ -492,17 +497,49 @@ jQuery(document).ready(function() {
...
+                if (paymentMethodType === 'paypal') {
+                    return true;
+                } else {
+                    resolve();
+                }
             }
         }
 
+        function handleOnCancel(state) {
+            hidePopup();
```
cyattilakiss, line 521: do we need to hidePopup() ?

I think it can be removed but pls double check it if it's used anywhere

```suggestion
```

views/js/checkout-component-renderer.js
```diff
@@ -492,17 +497,49 @@ jQuery(document).ready(function() {
...
-                resolve();
+                if (paymentMethodType === 'paypal') {
+                    return true;
+                } else {
+                    resolve();
+                }
             }
         }
 
+        function handleOnCancel(state) {
```
cyattilakiss, line 520: Doesn't the component exist in this callback here as a parameter?
I think you might not need to retrieve it with this `const component = getFormComponent(this.paymentForm);`

```suggestion
        function handleOnCancel(state, component) {
```

views/js/checkout-component-renderer.js
```diff
@@ -492,17 +497,49 @@ jQuery(document).ready(function() {
...
+            hidePopup();
+            const component = getFormComponent(this.paymentForm);
+            processPaymentsDetails({
+                'details': { 'orderID': state.orderID },
```
cyattilakiss, line 524: does the state have the orderID?  Shouldn't this be self?

controllers/FrontController.php
```diff
@@ -170,9 +170,10 @@ protected function getCurrentCart()
      * @param $cart
      * @param $customer
      * @param $isAjax
+     * @param $cancelled
      * @throws AdyenException
      */
-    protected function handlePaymentsResponse($response, $cart, $customer, $isAjax)
+    protected function handlePaymentsResponse($response, $cart, $customer, $isAjax, $cancelled = false)
```
cyattilakiss, line 176: Please double check if the paypal payments/details response is going to be indeed Cancelled or Refused, in case it's cancelled please just use the resultCode and you don't need the cancelled flag from the frontend

views/js/checkout-component-renderer.js
```diff
@@ -108,7 +108,8 @@ jQuery(document).ready(function() {
         var placeOrderAllowed;
         var popupModal;
 
-        var notSupportedComponents = ['paypal', 'giropay'];
+        var notSupportedComponents = ['giropay'];
```
cyattilakiss, line 111: pls rename it to skipComponents

controllers/FrontController.php
```diff
@@ -222,19 +223,26 @@ protected function handlePaymentsResponse($response, $cart, $customer, $isAjax)
...
                 $this->adyenPaymentResponseModel->deletePaymentResponseByCartId($cart->id);
 
                 // In case of refused payment there is no order created and the cart needs to be cloned and reinitiated
                 $this->cartService->cloneCurrentCart($this->context, $cart);
                 $this->logger->error("The payment was refused, id:  " . $cart->id);
 
+                if ($cancelled) {
+                    $message = $this->l('The payment was cancelled by the customer');
+                } else {
+                    $message = $this->l('The payment was refused');
```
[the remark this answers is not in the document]
Morerice: https://devdocs.prestashop.com/1.7/modules/creation/module-translation/classic-system/

acampos1916 (dismissed):

controllers/FrontController.php
```diff
@@ -170,6 +170,7 @@ protected function getCurrentCart()
      * @param $cart
      * @param $customer
      * @param $isAjax
+     * @param $cancelled
```
cyattilakiss, line 173:
```suggestion
```

controllers/FrontController.php
```diff
@@ -222,19 +223,27 @@ protected function handlePaymentsResponse($response, $cart, $customer, $isAjax)
...
+            case 'Cancelled':
                 // PaymentResponse can be deleted
                 $this->adyenPaymentResponseModel->deletePaymentResponseByCartId($cart->id);
 
-                // In case of refused payment there is no order created and the cart needs to be cloned and reinitiated
+                // In case of refused/cancelled payment there is no order created and the cart needs to be cloned and
+                // reinitiated
                 $this->cartService->cloneCurrentCart($this->context, $cart);
-                $this->logger->error("The payment was refused, id:  " . $cart->id);
+                $this->logger->error('The payment was ' . strtolower($resultCode) . ', id:  ' . $cart->id);
```
cyattilakiss, line 233:
```suggestion
                $this->logger->error('The payment was ' . strtolower($resultCode) . ', with cart id:  ' . $cart->id);
```

cyattilakiss (approved):

acampos1916 (approved):

PR #156, opened 2021-02-08
[PW-3960] - Verify the paid amount during notification processing 

Morerice:
## Summary
When processing authorisation w/ `success = true`, verify that the notification amount, order and cart amounts are all equal. If one of these fields does not match, flag it with the `ADYEN_OS_PAYMENT_NEEDS_ATTENTION` status.

## Tested scenarios
* Manually updated notification amount after order was completed
* Manually updated order amount after order was completed

### DIFF
--- a/service/notification/NotificationProcessor.php
+++ b/service/notification/NotificationProcessor.php
@@ -29,8 +29,10 @@ use Adyen\PrestaShop\model\AdyenNotification;
 use Adyen\PrestaShop\model\AdyenPaymentResponse;
 use Adyen\PrestaShop\service\adapter\classes\CustomerThreadAdapter;
 use Adyen\PrestaShop\service\adapter\classes\order\OrderAdapter;
+use Adyen\Util\Currency;
 use Context;
 use Db;
+use OrderCore;
 use PrestaShopDatabaseException;
 use PrestaShopException;
 use Psr\Log\LoggerInterface;
@@ -92,6 +94,11 @@ class NotificationProcessor
     private $orderService;
 
     /**
+     * @var Currency
+     */
+    private $utilCurrency;
+
+    /**
      * NotificationProcessor constructor.
      *
      * @param AdyenHelper $helperData
@@ -101,6 +108,8 @@ class NotificationProcessor
      * @param LoggerInterface $logger
      * @param Context $context
      * @param AdyenPaymentResponse $adyenPaymentResponse
+     * @param OrderService $orderService
+     * @param Currency $utilCurrency
      */
     public function __construct(
         AdyenHelper $helperData,
@@ -110,7 +119,8 @@ class NotificationProcessor
         LoggerInterface $logger,
         Context $context,
         AdyenPaymentResponse $adyenPaymentResponse,
-        OrderService $orderService
+        OrderService $orderService,
+        Currency $utilCurrency
     ) {
         $this->helperData = $helperData;
         $this->dbInstance = $dbInstance;
@@ -120,11 +130,13 @@ class NotificationProcessor
         $this->context = $context;
         $this->adyenPaymentResponse = $adyenPaymentResponse;
         $this->orderService = $orderService;
+        $this->utilCurrency = $utilCurrency;
     }
 
     /**
      * @param $unprocessedNotification
      * @return bool
+     * @throws \Exception
      */
     public function processNotification($unprocessedNotification)
     {
@@ -165,6 +177,10 @@ class NotificationProcessor
                         // Add additional data to order if there is any (only possible when the notification success is
                         // true
                         $this->orderService->addPaymentDataToOrderFromResponse($order, $unprocessedNotification);
+                    }
+
+                    if (!$this->validateWithCartAndOrder($unprocessedNotification, $order)) {
+                        $order->setCurrentState(\Configuration::get('ADYEN_OS_PAYMENT_NEEDS_ATTENTION'));
                     }
 
                     // In case psp reference is missing from the order_payment add it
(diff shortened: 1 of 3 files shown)


service/notification/NotificationProcessor.php
```diff
@@ -335,4 +351,32 @@ private function isCurrentOrderStatusANonFinalStatus($currentOrderStatus)
...
+    /**
+     * @param $notification
+     * @param OrderCore $order
+     * @return bool
+     * @throws \Exception
+     */
+    private function validateWithCartAndOrder($notification, OrderCore $order)
+    {
+        $cart = \Cart::getCartByOrderId($order->id);
+        $cartCurrency = \Currency::getCurrency($cart->id_currency);
```
cyattilakiss, line 364: Do the same with the order currency

```suggestion
        $cartCurrency = \Currency::getCurrency($cart->id_currency);
        $orderCurrency = \Currency::getCurrency($order->id_currency);
        $orderCurrencyIso = $orderCurrency['iso_code'];
```

service/notification/NotificationProcessor.php
```diff
@@ -335,4 +351,32 @@ private function isCurrentOrderStatusANonFinalStatus($currentOrderStatus)
...
+        $cartCurrencyIso = $cartCurrency['iso_code'];
+
+        $cartTotalMinorUnits = $this->utilCurrency->sanitize($cart->getOrderTotal(), $cartCurrencyIso);
+        $orderTotalMinorUnits = $this->utilCurrency->sanitize($order->getTotalPaid(), $cartCurrencyIso);
```
cyattilakiss, line 368:
```suggestion
        $orderTotalMinorUnits = $this->utilCurrency->sanitize($order->getTotalPaid(), $orderCurrencyIso);
```

service/notification/NotificationProcessor.php
```diff
@@ -335,4 +351,32 @@ private function isCurrentOrderStatusANonFinalStatus($currentOrderStatus)
...
+
+    /**
+     * @param $notification
+     * @param OrderCore $order
+     * @return bool
+     * @throws \Exception
+     */
+    private function validateWithCartAndOrder($notification, OrderCore $order)
+    {
+        $cart = \Cart::getCartByOrderId($order->id);
```
cyattilakiss, line 363: Whenever you work with prestashop core objects you can validate thembefore using them with the \Validate::isLoadedObject() function

service/notification/NotificationProcessor.php
```diff
@@ -335,4 +351,32 @@ private function isCurrentOrderStatusANonFinalStatus($currentOrderStatus)
...
+        $cartCurrency = \Currency::getCurrency($cart->id_currency);
+        $cartCurrencyIso = $cartCurrency['iso_code'];
+
+        $cartTotalMinorUnits = $this->utilCurrency->sanitize($cart->getOrderTotal(), $cartCurrencyIso);
+        $orderTotalMinorUnits = $this->utilCurrency->sanitize($order->getTotalPaid(), $cartCurrencyIso);
+
+        if ($notification['amount_currency'] !== $cartCurrencyIso ||
```
cyattilakiss, line 370:
```suggestion
        if ($notification['amount_currency'] !== $cartCurrencyIso || $notification['amount_currency'] !== $orderCurrencyIso ||
```

service/notification/NotificationProcessor.php
```diff
@@ -167,6 +179,10 @@ public function processNotification($unprocessedNotification)
                         $this->orderService->addPaymentDataToOrderFromResponse($order, $unprocessedNotification);
                     }
 
+                    if (!$this->validateWithCartAndOrder($unprocessedNotification, $order)) {
```
cyattilakiss, line 182: this should be done before setting the order state to PS_OS_PAYMENT - in case it's set to PS_OS_PAYMENT at first then the customer is going to get an email that they payment was successful while after it's tagged as needs attention

service/notification/NotificationProcessor.php
```diff
@@ -156,10 +168,13 @@ public function processNotification($unprocessedNotification)
         // Process notifications based on it's event code
         switch ($unprocessedNotification['event_code']) {
             case AdyenNotification::AUTHORISATION:
-                // Notification success is 'true'
                 if ('true' === $unprocessedNotification['success']) {
-                    // Moves order to paid if order status is not paid already
-                    if ($this->isCurrentOrderStatusANonFinalStatus($order->getCurrentState())) {
+                    // If notification data does not match cart and order, set to PAYMENT_NEEDS_ATTENTION
+                    // Else if not in a final status, set to PAYMENT
+                    if (!$this->validateWithCartAndOrder($unprocessedNotification, $order)) {
```
cyattilakiss, line 174: you can also do an early return since the notification is processed and the process is finished.
The elseif increasis the cognitive complexity and it makes it hard to understand the business logic
It's easier just to read like:

1. validate it - if it's not valid update order status and finish the process
2. update order status if ...
3. ...

service/notification/NotificationProcessor.php
```diff
@@ -335,4 +350,40 @@ private function isCurrentOrderStatusANonFinalStatus($currentOrderStatus)
...
+     * @param OrderCore $order
+     * @return bool
+     * @throws \Exception
+     */
+    private function validateWithCartAndOrder($notification, OrderCore $order)
+    {
+        $cart = \Cart::getCartByOrderId($order->id);
+
+        if (!\Validate::isLoadedObject($cart)) {
+            return false;
```
cyattilakiss, line 365: pls add a log here so we know what went wrong and not the actual validation failed

cyattilakiss (approved):

service/notification/NotificationProcessor.php
```diff
@@ -335,4 +354,48 @@ private function isCurrentOrderStatusANonFinalStatus($currentOrderStatus)
...
+                )
+            );
+
+            return false;
+        }
+
+        $cartCurrency = \Currency::getCurrency($cart->id_currency);
+        $orderCurrency = \Currency::getCurrency($order->id_currency);
+        $cartCurrencyIso = $cartCurrency['iso_code'];
+        $orderCurrencyIso = $orderCurrency['iso_code'];
```
acampos1916, line 383: Are we certain that these arrays will contain those indexes or should we do some defensive programming here with `!empty()`?

acampos1916 (commented):
Looks good, see comment

service/notification/NotificationProcessor.php
cyattilakiss: It's like an object that should always have a value - if not then it's more of a core issue in the shop. Of course we can add some defensive programming, although in that case I'd create an adapter for the object with a function which always does the validation before retrieving the property

acampos1916 (approved):

PR #157, opened 2021-02-09
[PW-4033] - Fix translations everywhere in the plugin where it's incorrect 

Morerice:
## Summary

1. Search for ->l( functions in the codebase and check if the translation is used correctly.
3. Fix the usage if incorrect
4. Check if the translation shows up in the admin where the translation can be added

**Invalid translation definition:**
- Any translation containing a variable may [yield unexpected results](https://devdocs.prestashop.com/1.7/modules/creation/module-translation/classic-system/#limitations-and-caveats)
- Any translation [containing a new line](https://github.com/PrestaShop/PrestaShop/issues/11573) between the opening bracket and the first quote will not be translatable
- Any translation containing a double quoted string ie. $this->l("Test string")will not be translatable

### DIFF
--- a/adyenofficial.php
+++ b/adyenofficial.php
@@ -1068,7 +1068,8 @@ class AdyenOfficial extends PaymentModule
                 $oneClickOption->setCallToActionText(
                     sprintf(
                         $this->l('Pay by saved %s ending: %s'),
-                        $storedPaymentMethod['name'], $storedPaymentMethod['lastFour']
+                        $storedPaymentMethod['name'],
+                        $storedPaymentMethod['lastFour']
                     )
                 )
                     ->setForm(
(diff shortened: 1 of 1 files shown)


adyenofficial.php
```diff
@@ -1099,8 +1066,10 @@ public function hookPaymentOptions()
 
                 $oneClickOption = new PrestaShop\PrestaShop\Core\Payment\PaymentOption();
                 $oneClickOption->setCallToActionText(
-                    $this->l(
-                        'Pay by saved ' . $storedPaymentMethod['name'] . " ending: " . $storedPaymentMethod['lastFour']
+                    sprintf(
+                        $this->l('Pay by saved %s ending: %s'),
```
cyattilakiss, line 1070: in case the translation adds one more %s then the function will show a php warning - I think it's better to break this translation up to 2 translations and then concatenating the text later

cyattilakiss (approved):

acampos1916 (approved):

PR #159, opened 2021-02-10
[PW-2374] -  Do not send shopperReference for guest users in the /paymentMethods request

Morerice:
## Summary
The current prestashop plugin sends the `shopperReference` for both logged in and guest users, since the platform automatically creates a customer for guest users. Hence, we should check if the customer is a guest user and if so, **do not send** the `shopperReference` parameter. Otherwise, this parameter should be normally sent.

## Tested scenarios
- Logged in scenario
- Guest user scenario

### DIFF
--- a/helper/Data.php
+++ b/helper/Data.php
@@ -114,7 +114,9 @@ class Data
             );
 
             return array();
-        } elseif (!\Validate::isLoadedObject($customer)) {
+        }
+
+        if (!\Validate::isLoadedObject($customer)) {
             $this->logger->error(
                 sprintf('Unable to load customer linked to Cart (%s)', $cart->id)
             );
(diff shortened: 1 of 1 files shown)


helper/Data.php
```diff
@@ -106,11 +106,19 @@ public function __construct(
...
         $merchantAccount = \Configuration::get('ADYEN_MERCHANT_ACCOUNT');
+        $customer = new \Customer($cart->id_customer);
 
         if (!$merchantAccount) {
             $this->logger->error(
                 "The merchant account field is empty, check your Adyen configuration in Prestashop."
             );
+
+            return array();
+        } elseif (!\Validate::isLoadedObject($customer)) {
```
cyattilakiss, line 117: Instead of an else if just do these validations in order and since you already do an early return it's easier to read the code.

```suggestion
        }
        
        if (!\Validate::isLoadedObject($customer)) {
```

cyattilakiss (approved):

acampos1916 (approved):

Issue #147, opened 2021-01-15
Shipping fees are refunded twice


YanP-Adyen:
**Describe the bug**
When an order with shipping fees is refunded, those fees are refunded twice.

Example : 1319056392467061

![image001-1](https://user-images.githubusercontent.com/68015043/104726741-705d2300-5734-11eb-9006-6e1bb6abc9a1.png)
See the lines before totals: 
Réductions = discount (not relevant here)
Livraison = shipping fees. (7,5EUR)
Total amount is 47,49EUR **with shipping fees included**

See logs attached, we send a refund request (4716101020123117) for 54,99EUR (which is = 47,49 + 7,5), obviously we reject it as the amount is bigger than the original payment. 
[extractLogs.log](https://github.com/Adyen/adyen-prestashop/files/5820429/extractLogs.log)

Morerice:
Hi @YanP-Adyen 

The issue occurred because we are currently adding the shipping amount to the credit slip whenever this is set, as can be seen here: https://github.com/Adyen/adyen-prestashop/blob/c4a20eb0b15445100fdcef1c5561307eb451a75c/service/modification/Refund.php#L91

Hence in this specific case the merchant specified the refund value to be **47,49EUR** in total, but because of this clause, the shipping cost was also automatically added to it, resulting in **54,99EUR**

Best regards,
Jeantwan

This should be fixed by checking the **shipping_cost** value in the OrderSlip entity. This value will be set whenever the merchant selects the option to also refund the shipping amount.

PR #163, opened 2021-02-19
[PW-4153] - Correctly check if shipping_cost should be added to OrderSlip

Morerice:
## Summary
Currently when an order slip is to be created, if the shipping_amount exists, it is always added to the OrderSlip amount:

```
            if (!empty($orderSlip->shipping_cost_amount)) {
                $fullRefundAmount += $orderSlip->shipping_cost_amount;
            }
```

Hence, instead of checking the amount, the `shipping_cost` flag should be set. This will be set whenever the merchant selects to also refund the shipping amount, from the admin interface.

Also, because of [this issue](https://github.com/Adyen/adyen-prestashop/issues/147) the amount sent to Adyen should be capped, to ensure that the merchant does not attempt to refund a larger amount than the customer actually paid. This issue has [already been identified](https://github.com/PrestaShop/PrestaShop/issues/18319) on prestashop side. This solution should be temporary so a TODO message to remove it once this is fixed, should also be added.

## Tested scenarios

* Refund on order with a discount and no shipping.
  * Exclude amount of initial voucher.
  * Refund amount = specified amount in admin interface = paid amount.
  * Correct credit slip
* Refund on order with a  discount and w/shipping.
  * Exclude amount of initial voucher and repay shipping.
  * Refund amount = specified amount in admin interface + shipping = paid amount.
  * Credit slip has incorrect values (prestashop issue)
* Refund on order with a discount and w/shipping.
  * Exclude amount of initial voucher and do not repay shipping.
  * Refund amount = specified amount in admin interface = paid amount.
  * Correct credit slip

### DIFF
--- a/service/modification/Refund.php
+++ b/service/modification/Refund.php
@@ -96,8 +96,8 @@ class Refund
 
         $fullRefundAmount = $orderSlip->amount;
 
-        // In case shipping cost amount is not empty add shipping costs to the order slip amount
-        if (!empty($orderSlip->shipping_cost)) {
+        // If admin wants to include shipping, add shipping costs to the order slip amount
+        if ($orderSlip->shipping_cost !== '0') {
             $fullRefundAmount += $orderSlip->shipping_cost_amount;
         }
 
--- a/service/modification/Refund.php
+++ b/service/modification/Refund.php
@@ -26,6 +26,7 @@ namespace Adyen\PrestaShop\service\modification;
 
 use Adyen\AdyenException;
 use Adyen\PrestaShop\infra\NotificationRetriever;
+use Adyen\PrestaShop\service\adapter\classes\order\OrderAdapter;
 use Adyen\PrestaShop\service\modification\exception\NotificationNotFoundException;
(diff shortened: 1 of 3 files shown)


acampos1916 (approved):

service/modification/Refund.php
```diff
@@ -96,8 +96,8 @@ public function request(OrderSlip $orderSlip, $currency)
 
         $fullRefundAmount = $orderSlip->amount;
 
-        // In case shipping cost amount is not empty add shipping costs to the order slip amount
-        if (!empty($orderSlip->shipping_cost)) {
+        // If admin wants to include shipping, add shipping costs to the order slip amount
+        if ($orderSlip->shipping_cost !== '0') {
```
cyattilakiss, line 100: I would check for the positive case here instead of anything else than string 0, jsut to be sure to add the shipping cost when it is "enabled" = 1

```suggestion
        if ($orderSlip->shipping_cost === '1') {
```

cyattilakiss (approved):

acampos1916 (approved):

PR #164, opened 2021-02-25
Fix error message in case of missing live url prefix

cyattilakiss:
## Summary
In case the mode is production and the live url prefix is empty the module shows an exception which is thrown by the library. In order to avoid this certain scenario the live url prefix is required from now on.
In case an (network or other non Adyen related) error shows up during the payment methods request we catch all these to avoid showing it on the frontend and instead we just log it in the error.log file in the adyen folder.

### DIFF
--- a/adyenofficial.php
+++ b/adyenofficial.php
@@ -565,7 +565,7 @@ class AdyenOfficial extends PaymentModule
             }
 
             if (empty($live_endpoint_url_prefix)) {
-                $output .= $this->displayError($this->l('Invalid Configuration value for live endpoint url prefix'));
+                $output .= $this->displayError($this->l('Invalid Configuration value for live endpoint URL prefix'));
             }
 
             if ($output == null) {
--- a/adyenofficial.php
+++ b/adyenofficial.php
@@ -564,6 +564,10 @@ class AdyenOfficial extends PaymentModule
                 $output .= $this->displayError($this->l('Invalid Configuration value for Notification Username'));
             }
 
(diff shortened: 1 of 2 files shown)


adyenofficial.php
```diff
@@ -564,6 +564,10 @@ public function getContent()
                 $output .= $this->displayError($this->l('Invalid Configuration value for Notification Username'));
             }
 
+            if (empty($live_endpoint_url_prefix)) {
+                $output .= $this->displayError($this->l('Invalid Configuration value for live endpoint url prefix'));
```
acampos1916, line 568:
```suggestion
                $output .= $this->displayError($this->l('Invalid Configuration value for live endpoint URL prefix'));
```

acampos1916 (approved):

helper/Data.php
```diff
@@ -154,7 +154,7 @@ public function fetchPaymentMethods(Cart $cart, $language)
         $responseData = "";
         try {
             $responseData = $this->adyenCheckoutService->paymentMethods($adyenFields);
-        } catch (AdyenException $e) {
+        } catch (\Exception $e) {
```
Morerice, line 157: Should we maybe keep the AdyenException catch block and then add the parent catch block (instead of replacing) so that we can have a different message for the two cases in order to easily identify if the issue is ours or not?

cyattilakiss: yes it is indeed a good idea :) Although in this case since there is no different logic per exception type I choose to handle all of them in the same catch which here only means logging it.

Morerice (approved):

PR #166, opened 2021-03-02
[PW-4243] - Do not remove paymentRow on 1.6

Morerice:
## Summary
Since #149 we started removing the payment method if isAvailable is false, both on 1.6 and 1.7:

```
.catch(e => {
                        if (IS_PRESTA_SHOP_16) {
                            const paymentRow = paymentForm.closest('.adyen-payment');
                            paymentRow.remove();
                        } else {
                            const payWithOption = paymentForm.closest('.js-payment-option-form');
                            const paymentOption = payWithOption.prev();
                            paymentOption.remove();
                            payWithOption.remove();
                        }
                    });
```

However on 1.6, the mutationObserver below will call the `renderPaymentMethods` whenever a mutation is observed, such as when a payment method is removed.

```
        // Callback function to execute when mutations are observed
        const callback = function(mutationsList, observer) {
            // Use traditional 'for loops' for IE 11
            for (const mutation of mutationsList) {
                if (mutation.type === 'childList') {
                    // The children are being changed so disconnet the observer
                    // at first to avoid infinite loop
                    observer.disconnect();
                    // Render the adyen checkout components
                    renderPaymentMethods();
                }
            }

            // Connect the observer again in case the checkbox is clicked
            // multiple times
            observer.observe(targetNode, config);
        };
```
Hence, this will cause issues with payment methods which must keep the same instance alive throughout, such as paypal.
This issue does not occur on 1.7, since this observer is not used. We should replace the remove() call with a hide() call in order to not trigger the observer.

## Tested scenarios
* PayPal flow on 1.7
* PayPal flow on 1.6

### DIFF
--- a/adyenofficial.php
+++ b/adyenofficial.php
@@ -1118,7 +1118,7 @@ class AdyenOfficial extends PaymentModule
                 }
 
                 $localPaymentMethod = new PrestaShop\PrestaShop\Core\Payment\PaymentOption();
-                $localPaymentMethod->setCallToActionText($this->l('Pay by ' . $paymentMethod['name']))
+                $localPaymentMethod->setCallToActionText(sprintf($this->l('Pay by %s'), $paymentMethod['name']))
                     ->setForm(
                         $this->context->smarty->fetch(
                             _PS_MODULE_DIR_ . $this->name .
--- a/views/templates/front/payment-method.tpl
+++ b/views/templates/front/payment-method.tpl
@@ -30,7 +30,9 @@
          data-local-payment-method="{$paymentMethodType|escape:'html':'UTF-8'}"
     >
         <div class="col-xs-12 col-md-6">
-            <div class="adyen-payment-method-label">{l s='Pay with ' mod='adyenofficial'}{$paymentMethodName|escape:'html':'UTF-8'}</div>
+            <div class="adyen-payment-method-label">
+                {l s='Pay with %s' sprintf=[{$paymentMethodName|escape:'html':'UTF-8'}] mod='adyenofficial'}
+            </div>
             <form action="{$paymentProcessUrl|escape:'html':'UTF-8'}"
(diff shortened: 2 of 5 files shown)


acampos1916 (dismissed):

adyenofficial.php
```diff
@@ -1118,7 +1118,7 @@ public function hookPaymentOptions()
                 }
 
                 $localPaymentMethod = new PrestaShop\PrestaShop\Core\Payment\PaymentOption();
-                $localPaymentMethod->setCallToActionText($this->l('Pay by ' . $paymentMethod['name']))
+                $localPaymentMethod->setCallToActionText(sprintf($this->l('Pay by %s'), $paymentMethod['name']))
```
cyattilakiss, line 1121: can you pls also add a suggestion to docs to let our merchants know how to use the %s in their translations? Is there any validation that we need to do before using $this->l('Pay by %s') in  sprintf() ? What happens if they put an extra %s in the translation like Pay by %s %s ?

views/templates/front/payment-method.tpl
```diff
@@ -30,7 +30,9 @@
          data-local-payment-method="{$paymentMethodType|escape:'html':'UTF-8'}"
     >
         <div class="col-xs-12 col-md-6">
-            <div class="adyen-payment-method-label">{l s='Pay with ' mod='adyenofficial'}{$paymentMethodName|escape:'html':'UTF-8'}</div>
+            <div class="adyen-payment-method-label">
+                {l s='Pay with %s' sprintf=[{$paymentMethodName|escape:'html':'UTF-8'}] mod='adyenofficial'}
```
cyattilakiss, line 34: same question applies here

cyattilakiss (approved):

acampos1916 (approved):

adyenofficial.php
Morerice: Docs suggestion: https://docs.adyen.com/plugins/prestashop#set-up-translations

PR #169, opened 2021-03-11
[PW 2778] - Display the payment methods as the default payment methods 

Morerice:
## Summary
Copy the default html structure of a default payment method when displaying our payment methods.
It helps our merchants with the customization, templates and styling.

## Tested scenarios
* Payment methods clearly shown on 1.7
Screen Shot 2021-03-11 at 9 28 58 AM

* Payment methods clearly shown on 1.6
Screen Shot 2021-03-11 at 9 29 43 AM

### DIFF
--- a/adyenofficial.php
+++ b/adyenofficial.php
@@ -466,7 +466,8 @@ class AdyenOfficial extends PaymentModule
             'ADYEN_APPLE_PAY_MERCHANT_NAME',
             'ADYEN_APPLE_PAY_MERCHANT_IDENTIFIER',
             'ADYEN_GOOGLE_PAY_GATEWAY_MERCHANT_ID',
-            'ADYEN_GOOGLE_PAY_MERCHANT_IDENTIFIER'
+            'ADYEN_GOOGLE_PAY_MERCHANT_IDENTIFIER',
+            'ADYEN_REMOVE_PAYMENT_DISPLAY_COLLAPSE'
         );
 
         $result = true;
@@ -550,6 +551,8 @@ class AdyenOfficial extends PaymentModule
             $apple_pay_merchant_identifier = Tools::getValue('ADYEN_APPLE_PAY_MERCHANT_IDENTIFIER');
             $google_pay_gateway_merchant_id = Tools::getValue('ADYEN_GOOGLE_PAY_GATEWAY_MERCHANT_ID');
             $google_pay_merchant_identifier = Tools::getValue('ADYEN_GOOGLE_PAY_MERCHANT_IDENTIFIER');
+            $remove_payment_display_collapse = Tools::getValue('ADYEN_REMOVE_PAYMENT_DISPLAY_COLLAPSE');
+
 
             // validating the input
             if (empty($merchant_account) || !Validate::isGenericName($merchant_account)) {
@@ -588,6 +591,7 @@ class AdyenOfficial extends PaymentModule
                 Configuration::updateValue('ADYEN_APPLE_PAY_MERCHANT_IDENTIFIER', $apple_pay_merchant_identifier);
                 Configuration::updateValue('ADYEN_GOOGLE_PAY_GATEWAY_MERCHANT_ID', $google_pay_gateway_merchant_id);
                 Configuration::updateValue('ADYEN_GOOGLE_PAY_MERCHANT_IDENTIFIER', $google_pay_merchant_identifier);
+                Configuration::updateValue('ADYEN_REMOVE_PAYMENT_DISPLAY_COLLAPSE', $remove_payment_display_collapse);
 
                 if (!empty($notification_password)) {
                     Configuration::updateValue('ADYEN_NOTI_PASSWORD', $this->crypto->encrypt($notification_password));
@@ -944,6 +948,31 @@ class AdyenOfficial extends PaymentModule
             'hint' => $this->l('')
         );
 
+        if ($this->versionChecker->isPrestaShop16()) {
+            $fields_form[1]['form']['input'][] = array(
+                'type' => 'radio',
+                'label' => $this->l('Remove collapsable payment display'),
+                'name' => 'ADYEN_REMOVE_PAYMENT_DISPLAY_COLLAPSE',
+                'values' => array(
+                    array(
+                        'id' => 'enable',
+                        'value' => 1,
+                        'label' => $this->l('Yes')
+                    ),
+                    array(
+                        'id' => 'disable',
+                        'value' => 0,
+                        'label' => $this->l('No')
+                    )
+                ),
+                'is_bool' => true,
+                // phpcs:ignore Generic.Files.LineLength.TooLong
+                'hint' => 'Indicates whether the payment methods should be rendered in a list of collapsable items,
+                 during checkout',
+                'required' => true
+            );
+        }
+
         $fields_form[2]['form'] = array(
             'legend' => array(
                 'title' => $this->l('Developer settings'),
@@ -1015,6 +1044,7 @@ class AdyenOfficial extends PaymentModule
             $apple_pay_merchant_identifier = Tools::getValue('ADYEN_APPLE_PAY_MERCHANT_IDENTIFIER');
             $google_pay_gateway_merchant_id = Tools::getValue('ADYEN_GOOGLE_PAY_GATEWAY_MERCHANT_ID');
             $google_pay_merchant_identifier = Tools::getValue('ADYEN_GOOGLE_PAY_MERCHANT_IDENTIFIER');
+            $remove_payment_display_collapse = Tools::getValue('ADYEN_REMOVE_PAYMENT_DISPLAY_COLLAPSE');
         } else {
             $merchant_account = Configuration::get('ADYEN_MERCHANT_ACCOUNT');
             $integrator_name = Configuration::get('ADYEN_INTEGRATOR_NAME');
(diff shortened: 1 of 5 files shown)


msilvagarcia (dismissed):

acampos1916 (dismissed):

cyattilakiss:
just in general would it be a good idea to introduce a feature toggle to enable/disable the frontend collapse functionality? I can imagine in case some merchants only use card payment methods they won't want to hide the card component but merchants with multiple methods they will.

views/js/checkout-component-renderer.js
```diff
@@ -634,4 +636,12 @@ jQuery(document).ready(function() {
     function isPlaceOrderInProgress() {
         return placeOrderInProgress;
     }
+
+    // When these buttons are clicked, do not close the payment method div
```
cyattilakiss, line 640: what do you think instead of listing the fields that won't trigger the collapse method if the collapse would only work for the payment name and the arrow like: 
image

And not the rest below. (Probably introducing some extra margin between the payment method title and the content would make it a bit more visible

views/templates/front/payment-method.tpl
```diff
@@ -27,26 +27,53 @@
...
     <div class="row adyen-payment {$paymentMethodType|escape:'html':'UTF-8'}"
-         data-local-payment-method="{$paymentMethodType|escape:'html':'UTF-8'}"
-    >
+         data-local-payment-method="{$paymentMethodType|escape:'html':'UTF-8'}">
+        {if $isPrestaShop16}
+        <div class="col-xs-12 col-md-12">
+            <div class="payment_module adyen-collapser collapsed" data-toggle="collapse"
+                 data-target="#collapse{$paymentMethodType|escape:'html':'UTF-8'}" aria-expanded="false"
+                 aria-controls="collapse{$paymentMethodType|escape:'html':'UTF-8'}"
+                 style="background-image: url(https://checkoutshopper-live.adyen.com/checkoutshopper/images/logos/medium/{$paymentMethodBrand|escape:'html':'UTF-8'}.png)">
```
cyattilakiss, line 36: could you please create a ticket to replace the burnt in image links to resources in the prestashop environment? Similar to the ticket: PW-3120

views/js/checkout-component-renderer.js
Morerice: Html structure updated

adyenofficial.php
```diff
@@ -466,7 +466,8 @@ private function removeConfigurationsFromDatabase()
             'ADYEN_APPLE_PAY_MERCHANT_NAME',
             'ADYEN_APPLE_PAY_MERCHANT_IDENTIFIER',
             'ADYEN_GOOGLE_PAY_GATEWAY_MERCHANT_ID',
-            'ADYEN_GOOGLE_PAY_MERCHANT_IDENTIFIER'
+            'ADYEN_GOOGLE_PAY_MERCHANT_IDENTIFIER',
+            'ADYEN_REMOVE_PAYMENT_DISPLAY_COLLAPSE'
```
cyattilakiss, line 470: because of the release strategy it would be easier to release a next version if the configuration was the other way around. Instead of removing the display collapse if it was enabling it.

In this case we can keep the default value as false when we release a new feature version so we won't break anything and in a next major release we can change it to by default true.

```suggestion
            'ADYEN_ENABLE_PAYMENT_DISPLAY_COLLAPSE'
```

adyenofficial.php
```diff
@@ -944,6 +948,31 @@ public function displayForm()
             'hint' => $this->l('')
         );
 
+        if ($this->versionChecker->isPrestaShop16()) {
+            $fields_form[1]['form']['input'][] = array(
+                'type' => 'radio',
+                'label' => $this->l('Remove collapsable payment display'),
+                'name' => 'ADYEN_REMOVE_PAYMENT_DISPLAY_COLLAPSE',
+                'values' => array(
```
cyattilakiss, line 956: is there a way to have a default value?

adyenofficial.php
```diff
@@ -1529,7 +1561,9 @@ private function getOneClickPaymentMethods(array $paymentMethods)
                 $smartyVariables = array(
                     'storedPaymentApiId' => $storedPayment['id'],
                     'name' => $storedPayment['name'],
-                    'number' => $storedPayment['lastFour']
+                    'logoBrand' => $storedPayment['brand'],
+                    'number' => $storedPayment['lastFour'],
+                    'removeCollapse' => \Configuration::get('ADYEN_REMOVE_PAYMENT_DISPLAY_COLLAPSE')
```
cyattilakiss, line 1566: In this case if the configuration cannot be found then there is no default value.
A quick solution can be to get the configuration before assigning the value to the smartyVariables array and check if the value is 1 (or true I'm not sure which one you are going to store in the db because of the `is_bool` field) otherwise you can assign 0 (or false) to it so you can be sure what you pass to the frontend is something you know for sure.

Longer solution: PW-1839 to have default value for both 1.6 and 1.7 (A branch is already prepared for it I'm not sure if anything is missing from the feature branch)

views/templates/front/payment-method.tpl
```diff
@@ -27,26 +27,52 @@
...
 {else}
     <div class="row adyen-payment {$paymentMethodType|escape:'html':'UTF-8'}"
-         data-local-payment-method="{$paymentMethodType|escape:'html':'UTF-8'}"
-    >
+         data-local-payment-method="{$paymentMethodType|escape:'html':'UTF-8'}">
+        {if $isPrestaShop16}
+        <div class="col-xs-12 col-md-12">
+            <div class="payment_module" style="background-image: url(https://checkoutshopper-live.adyen.com/checkoutshopper/images/logos/medium/{$paymentMethodBrand|escape:'html':'UTF-8'}.png)">
+                {* If collapsing is NOT disabled due to config *}
+                <span {if empty($removeCollapse)} class="adyen-payment-method-label adyen-collapser collapsed" data-toggle="collapse" data-target="#collapse{$paymentMethodType|escape:'html':'UTF-8'}" aria-expanded="false" aria-controls="collapse{$paymentMethodType|escape:'html':'UTF-8'}"
```
cyattilakiss, line 35: In case you know exactly what can be the 2 values of the $enableCollapse then it's easier to read if you are checking for that and not if it's empty.

```suggestion
                <span {if $enableCollapse === true} class="adyen-payment-method-label adyen-collapser collapsed" data-toggle="collapse" data-target="#collapse{$paymentMethodType|escape:'html':'UTF-8'}" aria-expanded="false" aria-controls="collapse{$paymentMethodType|escape:'html':'UTF-8'}"
```

views/templates/front/payment-method.tpl
```diff
@@ -27,26 +27,52 @@
...
+                        {else} class="adyen-payment-method-label" {/if}>
+                    {l s='Pay by %s' sprintf=[{$paymentMethodName|escape:'html':'UTF-8'}] mod='adyenofficial'}
+                </span>
+                <div class="row">
+                    <div class="col-xs-12 col-md-6">
+                        <form action="{$paymentProcessUrl|escape:'html':'UTF-8'}"
+                              class="adyen-payment-form-{$paymentMethodType|escape:'html':'UTF-8'} additional-information" method="post">
+                            {* If collapsing is NOT disabled due to config *}
+                            <div {if empty($removeCollapse)} id="collapse{$paymentMethodType|escape:'html':'UTF-8'}" class="adyen-collapsable collapse" {/if}>
```
cyattilakiss, line 44:
```suggestion
                            <div {if $enableCollapse === true} id="collapse{$paymentMethodType|escape:'html':'UTF-8'}" class="adyen-collapsable collapse" {/if}>
```

adyenofficial.php
Morerice: Based on http://doc.prestashop.com/display/PS16/Using+the+HelperForm+class#UsingtheHelperFormclass-Radiobutton and my tests, there doesn't seem to be a way of specifying a default value

views/templates/front/payment-method.tpl
Morerice: The empty should have been removed before the committed changes, since the checks work without it

adyenofficial.php
Morerice: Change committed

adyenofficial.php
Morerice: Change committed

adyenofficial.php
cyattilakiss: thanks for checking it! 👍

adyenofficial.php
```diff
@@ -944,6 +948,31 @@ public function displayForm()
...
+                        'id' => 'disable',
+                        'value' => 0,
+                        'label' => $this->l('Disable')
+                    )
+                ),
+                'is_bool' => true,
+                // phpcs:ignore Generic.Files.LineLength.TooLong
+                'hint' => 'Indicates whether the payment methods should be rendered in a list of collapsable items,
+                 during checkout',
+                'required' => true
```
cyattilakiss, line 972: We are using a default 0 basically in the code so I think it's not needed to give it a required flag.

```suggestion
                'required' => false
```

Also I only found to have a default value if you set the configuration value in the installation and in the upgrade script. It would mean to set the ADYEN_PAYMENT_DISPLAY_COLLAPSE config to 0

acampos1916 (approved):

acampos1916 (approved):

cyattilakiss (approved):

PR #170, opened 2021-03-11
[PW-4295] - Update payment needs attention check

Morerice:
## Summary
In the [check](https://github.com/Adyen/adyen-prestashop/blob/develop/service/notification/NotificationProcessor.php#L388-L397) which compares the amounts and currencies, we should make the following changes:

* Double check if order values should be validated as well or only cart details should

* Log the cart and order details (similar to `FrontController::validateCartOrderTotalAndCurrency`)

## Tested scenarios
* Process notification after normal payment on 1.7
* Process notification after 3DS payment on 1.7
* Process notification after 3DS payment on 1.6


acampos1916 (approved):

msilvagarcia (commented):
What do you think of splitting this with the context variable from the logger interface?

cyattilakiss (approved):

PR #172, opened 2021-03-18
[PW-2778] - Display the payment methods as the default payment methods (PR2)

Morerice:
## Summary
Small css updated related to PR #169

### DIFF
--- a/views/templates/front/payment-method.tpl
+++ b/views/templates/front/payment-method.tpl
@@ -30,7 +30,7 @@
          data-local-payment-method="{$paymentMethodType|escape:'html':'UTF-8'}">
         {if $isPrestaShop16}
         <div class="col-xs-12 col-md-12">
-            <div class="payment_module adyen-collapse" style="background-image: url(https://checkoutshopper-live.adyen.com/checkoutshopper/images/logos/medium/{$paymentMethodBrand|escape:'html':'UTF-8'}.png)">
+            <div class="payment_module adyen-collapse-styling" style="background-image: url(https://checkoutshopper-live.adyen.com/checkoutshopper/images/logos/medium/{$paymentMethodBrand|escape:'html':'UTF-8'}.png)">
                 {* If collapsing is enabled by config *}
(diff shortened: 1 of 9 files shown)


views/templates/front/payment-method.tpl
```diff
@@ -30,7 +30,7 @@
          data-local-payment-method="{$paymentMethodType|escape:'html':'UTF-8'}">
         {if $isPrestaShop16}
         <div class="col-xs-12 col-md-12">
-            <div class="payment_module" style="background-image: url(https://checkoutshopper-live.adyen.com/checkoutshopper/images/logos/medium/{$paymentMethodBrand|escape:'html':'UTF-8'}.png)">
+            <div class="payment_module adyen-collapse" style="background-image: url(https://checkoutshopper-live.adyen.com/checkoutshopper/images/logos/medium/{$paymentMethodBrand|escape:'html':'UTF-8'}.png)">
```
cyattilakiss, line 33: isn't it only added when `$collapsePayments`?

views/templates/front/stored-payment-method.tpl
```diff
@@ -30,7 +30,7 @@
          data-stored-payment-api-id="{$storedPaymentApiId|escape:'html':'UTF-8'}">
         {if $isPrestaShop16}
             <div class="col-xs-12 col-md-12">
-                <div class="payment_module" style="background-image: url(https://checkoutshopper-live.adyen.com/checkoutshopper/images/logos/medium/{$logoBrand|escape:'html':'UTF-8'}.png)">
+                <div class="payment_module adyen-collapse" style="background-image: url(https://checkoutshopper-live.adyen.com/checkoutshopper/images/logos/medium/{$logoBrand|escape:'html':'UTF-8'}.png)">
```
cyattilakiss, line 33: same here

views/templates/front/payment-method.tpl
Morerice: The new class is just used to apply 1.6 styling to it, not to denote if the actual div is collapsible or not. Maybe renaming it to `adyen-collapse-styling` would make it more clear?

cyattilakiss: @Morerice yes I think it would be a bit better indeed

cyattilakiss (approved):

acampos1916 (approved):

PR #173, opened 2021-03-18
[PW-4253] Check if order states are already installed when installing them

cyattilakiss:
## Summary
Update order state installation to be more robust

Refactor the createPaymentNeedsAttentionOrderStatus() and
createWaitingForPaymentOrderStatus() to be more failsafe in case of a
configuration or state removal directly from the database

## Tested scenarios
Simple reinstallation without touching the order states or the configurations in the database
Reinstallation with removing the order states or the configurations in the database


adyenofficial.php
```diff
@@ -366,31 +375,45 @@ public function createAdyenOrderStatuses()
...
-            $order_state->name = array();
+        $orderStateConfigurationId = Configuration::get('ADYEN_OS_WAITING_FOR_PAYMENT');
+
+        $orderState = false;
+        if ($orderStateConfigurationId) {
+            $orderState = $this->orderStateAdapter->getOrderStateById($orderStateConfigurationId);
+        }
+
+        // In case order state does not exist in the database anymore
+        if (empty($orderState)) {
```
acampos1916, line 386: I think this should also work? Just simplifying a bit more.
```suggestion
        if ($orderStateConfigurationId) {
            $orderState = $this->orderStateAdapter->getOrderStateById($orderStateConfigurationId);
        }

        // In case order state does not exist in the database anymore
        if (!empty($orderState)) {
```

acampos1916 (approved):

adyenofficial.php
cyattilakiss: We only want to create a new order state if the $orderState is empty.

The only thing that could be removed is the `$orderState = false;` declaration at the beginning, but I don't like to check undefined variables even though in php it would work.

msilvagarcia (approved):

Morerice (approved):

PR #177, opened 2021-03-23
[PW-4375] - Set pspReference on redirect flow 

Morerice:
## Summary
Currently after the 3DS redirect, the pspReference is not being synchronously set in the order_payment.transaction_id field.

It is only being set when the notification of that particular payment is processed. Hence, any refund attempt on a 3DS payment before its notification is processed will fail, since the pspReference will not be saved.

**Steps:**
1. Create new service
2. Add new function in this service which will be responsible to update this field IFF it is not already set
3. Use this function in the FrontController and the NotificationProcessor

## Tested scenarios
1. Non-3DS (no redirect) payment - `order_payment.transaction_id` immediately set
2. 3DS2 (modal) payment - `order_payment.transaction_id` immediately set
3. 3DS1 (redirect) payment - `order_payment.transaction_id` set after redirect is done
4. In redirect flow, if for some reason the customer is not successfully redirected, `order_payment.transaction_id` set when notification is processed
5. Refund successful before notification is processed


service/notification/NotificationProcessor.php
```diff
@@ -204,8 +224,7 @@ public function processNotification($unprocessedNotification)
                             // Moves order to canceled
                             $this->orderService->updateOrderState(
                                 $order,
-                                \Configuration::get('PS_OS_CANCELED'),
-                                $extraVars
```
cyattilakiss, line 208: why removing the extravars here? In case the auth false notification contains a pspreference we can still update the order payment with it, can't we?

Morerice: Originally I removed the `extraVars` from the `updateOrderState` function so that the pspReference is updated in another function to have better single responsibility. That being said, I should have called this function in this specific case, which was added in this [commit](https://github.com/Adyen/adyen-prestashop/pull/177/commits/6bf71230405a7ffd227b4e00d386976e201e0526)

service/OrderPaymentService.php
```diff
@@ -0,0 +1,73 @@
...
+     * @param OrderPayment $orderPayment
+     * @param string $pspReference
+     *
+     * @return OrderPayment
+     * @throws \PrestaShopException
+     */
+    public function addPspReferenceForOrderPayment(OrderPayment $orderPayment, $pspReference)
+    {
+        if (\Validate::isLoadedObject($orderPayment)) {
+            if ($orderPayment !== false && empty($orderPayment->transaction_id)) {
```
cyattilakiss, line 46: The `$orderPayment !== false` is already validated by the \Validate::isLoadedObject($orderPayment)

```suggestion
            if (empty($orderPayment->transaction_id)) {
```

controllers/FrontController.php
```diff
@@ -475,14 +485,22 @@ protected static function getArrayOnlyWithApprovedKeys($array, $approvedKeys)
...
      */
     private function createOrUpdateOrder($cart, $extraVars, $customer, $orderStatus)
     {
         // Load order if exists from cart id
         if ($cart->OrderExists() !== false) {
             $order = $this->orderAdapter->getOrderByCartId($cart->id);
             if (\Validate::isLoadedObject($order)) {
-                $this->orderService->updateOrderState($order, $orderStatus, $extraVars);
+                $this->orderService->updateOrderState($order, $orderStatus);
+                $latestOrderPayment = $this->orderPaymentService->getLatestAdyenOrderPayment($order);
```
cyattilakiss, line 497: I am not sure about this so it's more like a question for this:

Does it make sense to use the latestAdyenPayment object here or maybe the first one makes more sense?
Since we create a new order for every redirect payment attempt and for the sync payments we only create the order when the payment status is not negative (rejected, cancelled, error) then theoretically we only should have 1 adyen payment object per order right?

If so, then I believe here we should update the first payment object instead of the last one.

Regarding its future proofness could you please check if refunds would create any extra order payment object for the order. If so then it's 100% sure that we should use the first payment object instead of the last one since then it would be another payment which we would update.

Could you please double check these concerns before merging it?

service/notification/NotificationProcessor.php
```diff
@@ -231,8 +242,7 @@ public function processNotification($unprocessedNotification)
                     if ($order->getCurrentState() === \Configuration::get('ADYEN_OS_WAITING_FOR_PAYMENT')) {
                         $this->orderService->updateOrderState(
                             $order,
-                            \Configuration::get('PS_OS_CANCELED'),
-                            $extraVars
+                            \Configuration::get('PS_OS_CANCELED')
                         );
```
cyattilakiss, line 246: If I'm not mistaken it's needed here as well.

```suggestion
                        );
                        
                        $this->setPspReferenceUsingNotificationData($order, $unprocessedNotification);
```

controllers/FrontController.php
Morerice: I'll check it, but I don't think Refunds create a new `order_payment`. So as mentioned we will probably have only 1 adyen order_payment per order

service/notification/NotificationProcessor.php
Morerice: Suggestion committed

controllers/FrontController.php
cyattilakiss: okay, let's create a follow up investigation ticket just to be sure we won't forget it and keep it as is for now 👍

cyattilakiss (dismissed):

controllers/FrontController.php
Morerice: First `order_payment` is now retrieved: https://github.com/Adyen/adyen-prestashop/pull/177/commits/57cacf292a72bc99996108e8136230896338f057

cyattilakiss (approved):

acampos1916 (approved):

PR #178, opened 2021-03-26
[PW-4366] Process the notifications in an asynchronous way

msilvagarcia:
## Summary

Add a convenient way to process notifications for the PrestaShop platform. If configured, it will open a socket to the proper host and make an HTTP request to the notification cron controller endpoint.

## Tested scenarios

- PrestaShop 1.6
- PrestaShop 1.7

**Fixed issue**:


infra/AsyncNotificationTrigger.php
```diff
@@ -0,0 +1,93 @@
...
+        }
+
+        $bytes = fwrite($socket, $request);
+        $this->logger->debug('Wrote to socket', [
+            'hostname' => $prefix . $endpointParts['host'],
+            'port' => $endpointParts['port'],
+            'request' => $request,
+            'bytes' => $bytes
+        ]);
+        usleep(16000);
```
cyattilakiss, line 83: why is this needed?

service/notification/NotificationReceiver.php
```diff
@@ -342,4 +349,23 @@ private function returnAccepted($acceptedMessage)
...
+
+    private function callAsyncCronRunner()
+    {
+        try {
+            /** @var Crypto $crypto */
+            $crypto = ServiceLocator::get('Adyen\PrestaShop\infra\Crypto');
+            /** @var AsyncNotificationTrigger $asyncNotificationTrigger */
+            $asyncNotificationTrigger = ServiceLocator::get('Adyen\PrestaShop\infra\AsyncNotificationTrigger');
+            $asyncNotificationTrigger->trigger(
+                $crypto->decrypt(\Configuration::get('ADYEN_ADMIN_PATH')),
```
cyattilakiss, line 361: isn't the `_PS_ADMIN_DIR_` available here as well instead of using a configuration for this?
I'm thinking if someone migrates to another server/hosting and the _PS_ADMIN_DIR_ changes then we won't update this value and the cron will fail to run.

msilvagarcia: It is not available here as this is a front-end controller

cyattilakiss: is it maybe a better idea instead of storing it encrypted in the db to create a service to retrieve this if it's not available in this class? I think it would cause some issues in the future because of the updates :(

msilvagarcia: It's difficult to get this because PrestaShop's admin URL is based on the file system and doesn't seem to be stored elsewhere. Also, shops cannot use the traditional "admin" and are told to rename the admin directory to something else the first time they try accessing it.

acampos1916 (approved):

Morerice (approved):

service/notification/NotificationReceiver.php
cyattilakiss: @msilvagarcia Yes indeed, I tried to find a workaround as well but it doesn't seem to be stored anywhere and it's just detected from the file system. Could you please create a follow up ticket to think about the next possible steps in order to avoid the future support tickets regarding this in case someone renames the admin or changes hosting therefore changing the path. A couple if ideas:
 - instead of an admin controller use a front controller for the notification processor
 - instead of storing the admin path as an encrypted value store it as a string so it still can be changed by the merchant
  - create a button that can regenerate the encrypted path config field in case of an error message that the notification processor cannot be found
  
Please add your ideas as well and let's refine it next week.

PR #184, opened 2021-04-02
Fix validation for payment methods

msilvagarcia:
## Summary

This fixes an issue in the Checkout page for the plugins. When one of the payment methods is set to valid, you could click the payment button or select a different payment method and finish the transaction without triggering the validation a second time.

This commit changes the flag that allows a payment to proceed to an object that holds states for different payment methods.

## Tested scenarios

For both PrestaShop 1.6 and 1.7:
- Switching payment methods and trying to pay
- Pay without switching payment methods

**Fixed issue**:

### DIFF
--- a/views/js/checkout-component-renderer.js
+++ b/views/js/checkout-component-renderer.js
@@ -15,7 +15,7 @@
  * Adyen PrestaShop plugin
  *
  * @author Adyen BV <support@adyen.com>
- * @copyright (c) 2020 Adyen B.V.
+ * @copyright (c) 2021 Adyen B.V.
  * @license https://opensource.org/licenses/MIT MIT license
  * This file is open source and available under the MIT license.
  * See the LICENSE file for more info.
@@ -128,7 +128,7 @@ jQuery(document).ready(function() {
 
     function renderPaymentMethods() {
         var selectedPaymentMethod;
-        var placeOrderAllowed;
+        var placeOrderAllowed = {};
         var popupModal;
 
         var skipComponents = ['giropay'];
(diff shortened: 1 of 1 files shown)


views/js/checkout-component-renderer.js
```diff
@@ -495,19 +495,19 @@ jQuery(document).ready(function() {
          * @param state
          */
         function handleOnChange(state) {
-            placeOrderAllowed = state.isValid;
+            placeOrderAllowed[state.data.paymentMethod.type] = state.isValid;
```
cyattilakiss, line 498: Is the `state.data.paymentMethod.type` prop available  for all the payment methods? Even if the method has no fields to render like klarna (simple redirect)

views/js/checkout-component-renderer.js
```diff
@@ -495,19 +495,19 @@ jQuery(document).ready(function() {
...
-            processPaymentsDetails(state.data).done(function(responseJSON) {
+            processPaymentsDetails(state.data).done(function (responseJSON) {
                 processControllerResponse(responseJSON,
                     getSelectedPaymentMethod());
             });
         }
 
-        function handleOnSubmit(state, component) {
-            placeOrderAllowed = state.isValid;
+        function handleOnSubmit(state/*, component*/) {
```
cyattilakiss, line 509: In case it's not used we can remove it.

```suggestion
        function handleOnSubmit(state) {
```

views/js/checkout-component-renderer.js
msilvagarcia: Payment methods without details won't trigger the `onChange` event anyways

cyattilakiss (approved):

Morerice (approved):

peterojo (approved):

PR #186, opened 2021-04-08
[PW-4199] - Use prestsahop logs page for easier log access

Morerice:
## Summary
We should log certain logs (depending on severity) using the prestashop functionality so that our merchants could get a better understanding about what's happening in the plugin when accessing the prestashop logs page. Only log the errors, warnings and major issues. Informative ones can still be just stored in the log files. `PrestaShopLoggerCore::addLog` function can be used to add logs to this page.

## Tested scenarios
* Error message logged
* Warning message logged

### DIFF
--- a/adyenofficial.php
+++ b/adyenofficial.php
@@ -958,7 +958,7 @@ class AdyenOfficial extends PaymentModule
                 'For configuration "ADYEN_APIKEY_LIVE" an exception was thrown: ' . $e->getMessage()
             );
         } catch (\Adyen\PrestaShop\exception\MissingDataException $e) {
-            $this->logger->error('The configuration "ADYEN_APIKEY_LIVE" has no value set.');
+            $this->logger->warning('The configuration "ADYEN_APIKEY_LIVE" has no value set.');
         }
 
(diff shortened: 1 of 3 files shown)


adyenofficial.php
```diff
@@ -958,7 +958,7 @@ public function displayForm()
                 'For configuration "ADYEN_APIKEY_LIVE" an exception was thrown: ' . $e->getMessage()
             );
         } catch (\Adyen\PrestaShop\exception\MissingDataException $e) {
-            $this->logger->debug('The configuration "ADYEN_APIKEY_LIVE" has no value set.');
+            $this->logger->error('The configuration "ADYEN_APIKEY_LIVE" has no value set.');
```
cyattilakiss, line 961: this can be a warning as well, we set this as a required field so the config can not be saved if it's not filled

cyattilakiss (approved):

adyenofficial.php
Morerice: Change committed

cyattilakiss (approved):

acampos1916 (approved):

PR #189, opened 2021-04-16
[PW-4260] - Create a validate button for module installation 

Morerice:
**Description:**

Validate if the database tables, order statuses and necessary configurations are existing or not. We see that the install function even if it returns false can end with install success which is incorrectly but handled by PrestaShop.

What we can do is to ask the merchants to click and validate their installation after installing the plugin.

**DB tables**
`ps_adyen_notification`
`ps_adyen_payment_response`

**Order states**
In the `ps_order_state` db table
With the IDs from the `ps_configuration` db table with the name of:
`ADYEN_OS_WAITING_FOR_PAYMENT`
`ADYEN_OS_PAYMENT_NEEDS_ATTENTION`

**Steps:**
1. Create a new tab for debugging purposes
2.  Add the button with the functionality of
 - Checking if the db tables above mentioned are existing
 - Checking if the configuration fields above mentioned (only technical like order status id and required fields) are present in the configuration db table
 - Checking if the statuses are actually present in the `ps_order_state` db table as well
 - Check f the installed hooks are indeed registered
3. Show error message if something if missing - reach out to tech support, or show success message if everything is in place

## Tested scenarios
* Validation on 1.6
* Validation on 1.6 (multi-site)
* Validation on 1.7
* Validation on 1.7 (multi-site)
* Updated E2E tests


controllers/admin/AdminAdyenOfficialPrestashopValidatorController.php
```diff
@@ -0,0 +1,286 @@
...
+    }
+
+    /**
+     * Get the url accessed when the button is clicked, to download the zip file
+     *
+     * @return string
+     */
+    private function getValidateUrl()
+    {
+        $adminUrl = Tools::getAdminUrl('admin-dev/index.php?controller=AdminAdyenOfficialPrestashopValidator&token=');
```
acampos1916, line 234: Is admin-dev variable and customizable? could it be different for some merchants?

helper/Data.php
```diff
@@ -336,4 +337,25 @@ public function getTemplateFromModulePath($templatePath)
...
+     * Execute basic SELECT statement. If table does not exist, catch exception and return false
+     *
+     * @param $tableName
+     * @return bool
+     *
+     */
+    public function tableExists($tableName)
+    {
+        $db = Db::getInstance();
+        $sql = 'SELECT * FROM ' . _DB_PREFIX_ . $tableName;
```
acampos1916, line 351: Maybe this could use a LIMIT 1 so that there's no unnecessary stress to fetch all rows. Or a different command like SHOW TABLES

Morerice: Function called eventually, automatically adds Limit 1:
`Db::getRow`:
```
    /**
     * Returns an associative array containing the first row of the query
     * This function automatically adds "LIMIT 1" to the query.
     *
     * @param string|DbQuery $sql the select query (without "LIMIT 1")
     * @param bool $use_cache Find it in cache first
     *
     * @return array|bool|object|null
     */
    public function getRow($sql, $use_cache = true)
```

controllers/admin/AdminAdyenOfficialPrestashopValidatorController.php
Morerice: Updated in https://github.com/Adyen/adyen-prestashop/pull/189/commits/d615c4512edabd35e5c5aad1829109242d0f72cf

acampos1916 (approved):

acampos1916 (approved):

msilvagarcia (approved):

acampos1916 (approved):

PR #193, opened 2021-05-07
[PW-4635] - One-page checkout issue on 1.6 

Morerice:
## Summary
On 1.6, if the merchant enables quick (one-page) checkout, the js in the checkout-component-renderer.js file is not executed at all.

The `ADYEN_CHECKOUT_CONFIG` is required for the **checkout-component-renderer** to function. Currently, this is set when the `hookDisplayPaymentTop` action is called.
However if one-page checkout is enabled, once the checkout page is loaded, `hookDisplayPaymentTop` is not executed, which implies that the `window.ADYEN_CHECKOUT_CONFIG` in the **checkout-component-renderer** will be set to null and so it will return without executing this js file.

Once the user fills in his guest checkout details and chooses the carrier, the payment options will be loaded, at which point the `hookDisplayPaymentTop` will be executed. However, since we do not call the **checkout-component-renderer** again (it uses `jQuery(document).ready()`), this file is not re-executed and so the functionality is unusable.

If the user refreshes, the guest user info is included in the session and the payment section of the checkout is displayed on page load. Hence, the `hookDisplayPaymentTop` will be executed and the **checkout-component-renderer** will be successfully loaded.

## Tested scenarios
Payment on 1.6 w/one-page checkout
Payment on 1.6 w/default checkout
Payment on 1.7
E2E tests

### DIFF
--- a/helper/Data.php
+++ b/helper/Data.php
@@ -28,6 +28,7 @@ use Address;
 use Adyen;
 use Adyen\AdyenException;
 use Adyen\Environment;
+use Adyen\PrestaShop\application\VersionChecker;
 use Adyen\PrestaShop\infra\Crypto;
 use Adyen\PrestaShop\service\adapter\classes\Configuration;
 use Adyen\PrestaShop\service\adapter\classes\Language;
@@ -73,6 +74,9 @@ class Data
      */
     private $crypto;
 
+    /** @var VersionChecker $versionChecker */
+    private $versionChecker;
+
     /**
      * Data constructor.
      *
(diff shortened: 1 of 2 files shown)


helper/Data.php
```diff
@@ -117,7 +124,7 @@ public function fetchPaymentMethods(Cart $cart, $language)
             return array();
         }
 
-        if (!\Validate::isLoadedObject($customer)) {
+        if (!\Validate::isLoadedObject($customer) && !$this->areGuestAndOnePageCheckout16()) {
```
cyattilakiss, line 127: not sure if I get this correctly.
Isn't the customer object required for both 1.6 and 1.7?

helper/Data.php
```diff
@@ -128,9 +135,16 @@ public function fetchPaymentMethods(Cart $cart, $language)
         $amount = $cart->getOrderTotal();
         $currencyData = Currency::getCurrency($cart->id_currency);
         $currency = $currencyData['iso_code'];
-        $address = new Address($cart->id_address_invoice);
-        $countryCode = Country::getIsoById($address->id_country);
         $shopperLocale = $this->languageAdapter->getLocaleCode($language);
+        $address = new Address($cart->id_address_invoice);
+
+        // If address is not created correctly and guest and one-page checkout are enabled, use context country
+        if (is_null($address->id_country) && $this->areGuestAndOnePageCheckout16()) {
```
cyattilakiss, line 142: does it depend on the `$this->areGuestAndOnePageCheckout16()` result or anly on the `$address` object if it does exist or not?

PR #202, opened 2021-06-02
[PW-4781] - Set addresses when cloning cart

Morerice:
## Summary
Currently, in the `Cart::cloneCart` function, we do not set the `cart->id_address_delivery` and the `cart->id_invoice_delivery`.

In some cases, when a failed transaction is redirected and if there are multiple addresses, these 2 address fields will not be correctly set and hence the user will have to restart checkout from the address page, with no info regarding the failed payment. If the customer only has a single address available, this is not an issue.

Hence we need to set these 2 fields when cloning the cart, so the customer is redirected back to the payments page.

**Steps:**
1. Set the `cart->id_address_delivery` and the `cart->id_invoice_delivery` fields in the `Cart::cloneCart` function.

## Tested scenarios
E2E tests

### DIFF
--- a/adyenofficial.php
+++ b/adyenofficial.php
@@ -235,7 +235,7 @@ class AdyenOfficial extends PaymentModule
                 'displayPaymentTop',
                 'actionFrontControllerSetMedia',
                 'paymentOptions',
-                'displayPaymentReturn',
+                'paymentReturn',
                 'actionOrderSlipAdd',
                 'actionEmailSendBefore'
             )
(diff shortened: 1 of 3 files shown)


acampos1916 (approved):

adyenofficial.php
```diff
@@ -235,7 +235,7 @@ public static function getAdyenHooks()
                 'displayPaymentTop',
                 'actionFrontControllerSetMedia',
                 'paymentOptions',
-                'displayPaymentReturn',
+                'paymentReturn',
```
cyattilakiss, line 238: nice catch!

cyattilakiss (approved):

PR #203, opened 2021-06-07
[PW-4695] - Create smaller sections in configuration panel 

Morerice:
## Summary
Split the current general configurations into three sections:

* Test parameters
* Live parameters
* General parameters

and only require validation on certain fields, based on the selected environment.

Also replace any mentions of **notification** with **webhook**.

## Tested scenarios
* Setting up testing env
* Setting up prod env
* E2E tests

Screen Shot 2021-05-27 at 2 49 49 PM

### DIFF
--- a/adyenofficial.php
+++ b/adyenofficial.php
@@ -853,15 +853,23 @@ class AdyenOfficial extends PaymentModule
                 $output .= $this->displayError($this->l('Invalid Configuration value for Notification Username'));
             }
 
-            if (empty($client_key_test)) {
+            if (empty($client_key_test) && $mode === 'test') {
                 $output .= $this->displayError($this->l('Invalid Configuration value for test client key'));
             }
 
-            if (empty($client_key_live)) {
+            if (empty($client_key_live) && $mode === 'live') {
                 $output .= $this->displayError($this->l('Invalid Configuration value for live client key'));
             }
 
-            if (empty($live_endpoint_url_prefix)) {
+            if (empty($api_key_test) && $mode === 'test') {
+                $output .= $this->displayError($this->l('Invalid Configuration value for test api key'));
+            }
+
+            if (empty($api_key_live) && $mode === 'live') {
+                $output .= $this->displayError($this->l('Invalid Configuration value for live api key'));
+            }
+
+            if (empty($live_endpoint_url_prefix) && $mode === 'live') {
                 $output .= $this->displayError($this->l('Invalid Configuration value for live endpoint URL prefix'));
             }
 
@@ -915,6 +923,7 @@ class AdyenOfficial extends PaymentModule
                 $output .= $this->displayConfirmation($this->l('Settings updated'));
             }
         }
+
         return $output . $this->displayGetStarted() . $this->displayForm();
     }
 
@@ -962,8 +971,12 @@ class AdyenOfficial extends PaymentModule
         // Get default Language
         $default_lang = (int)Configuration::get('PS_LANG_DEFAULT');
 
-        // Get current mode
-        $mode = Configuration::get('ADYEN_MODE');
+        // Get the mode before all the other params since it's needed to check which fields are required
+        if (Tools::isSubmit('submit' . $this->name)) {
+            $mode = (string)Tools::getValue('ADYEN_MODE');
+        } else {
+            $mode = Configuration::get('ADYEN_MODE');
+        }
 
         // Init Fields form array
         $fields_form = array(array());
@@ -1427,7 +1440,6 @@ class AdyenOfficial extends PaymentModule
             // get settings from post because post can give errors and you want to keep values
             $merchant_account = (string)Tools::getValue('ADYEN_MERCHANT_ACCOUNT');
             $integrator_name = (string)Tools::getValue('ADYEN_INTEGRATOR_NAME');
-            $mode = (string)Tools::getValue('ADYEN_MODE');
             $notification_username = (string)Tools::getValue('ADYEN_NOTI_USERNAME');
             $cron_job_token = $cronjobToken;
             $auto_cron_job_runner = Tools::getValue('ADYEN_AUTO_CRON_JOB_RUNNER');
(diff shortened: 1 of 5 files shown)


adyenofficial.php
```diff
@@ -185,11 +185,13 @@ public function __construct()
     }
 
     /**
+     * @param $mode
+     *
      * @return string[]
      */
-    public static function getAdyenConfigNames()
```
cyattilakiss, line 190: If I'm not mistaken this is a full list of all the configs and is used also when the plugin is being uninstalled. In case we do this change then based on the config live - test it will only remove part of the configurations from the db instead of all

adyenofficial.php
```diff
@@ -215,6 +212,17 @@ public static function getAdyenConfigNames()
             'ADYEN_ENABLE_STORED_PAYMENT_METHODS',
             'ADYEN_PAYMENT_DISPLAY_COLLAPSE'
         );
+
+        if ($mode === 'test') {
```
cyattilakiss, line 216: I think this check needs to be done [here](https://github.com/Adyen/adyen-prestashop/pull/203/files#diff-6fb035a8c069d2f2a8569fffb0e36036621d864e8e76d6ef65bb714253a51615R851) when validating the input fields

adyenofficial.php
```diff
@@ -1011,71 +1108,108 @@ public function displayForm()
             'required' => true
         );
```
cyattilakiss, line 1110: I'd separate the settings with comments just to make it a bit easier to read and understand the code later (also for the other sections)

```suggestion
// Production settings
```

adyenofficial.php
```diff
@@ -1011,71 +1108,108 @@ public function displayForm()
...
-            'label' => '<a target="_blank" href="https://docs.adyen.com/plugins/prestashop#generate-an-api-key">API key for Test</a>',
-            'name' => 'ADYEN_APIKEY_TEST',
-            'desc' => $apiKeyTestLastDigits ? $this->l('Saved key ends in: ') . $apiKeyTestLastDigits :
-                $this->l('Please fill your API key for Test'),
-            'class' => $apiKeyTestLastDigits ? 'adyen-input-green' : '',
+            'label' => '<a target="_blank" href="https://docs.adyen.com/plugins/prestashop#generate-an-api-key">API key for Live</a>',
+            'name' => 'ADYEN_APIKEY_LIVE',
+            'desc' => $apiKeyLiveLastDigits ? $this->l('Saved key ends in: ') . $apiKeyLiveLastDigits :
+                $this->l('Please fill your API key for Live'),
+            'class' => $apiKeyLiveLastDigits ? 'adyen-input-green' : '',
```
cyattilakiss, line 1132: Since we got some feedback for the green border color maybe we can change this to password and if the value is ***** then we just act like nothing is submitted. Therefore we can get rid of the green coloring, we still don't pass back the unencrypted API key and it's maybe easier for the merchants to understand the UI.

adyenofficial.php
Morerice: https://github.com/Adyen/adyen-prestashop/pull/203/commits/518f6c41b9dd8780fb9af5e666c67f2f5afcd24e

adyenofficial.php
Morerice: https://github.com/Adyen/adyen-prestashop/pull/203/commits/74365fa57356f72033997b7a74ea5128e7e68ec4

adyenofficial.php
Morerice: https://github.com/Adyen/adyen-prestashop/pull/203/commits/e46540447f73d84ddf9a31e84142a3e6caf8b815

adyenofficial.php
```diff
@@ -850,21 +872,43 @@ public function getContent()
...
-            if (empty($client_key_live)) {
+            if (empty($client_key_live) && $mode === 'live') {
                 $output .= $this->displayError($this->l('Invalid Configuration value for live client key'));
             }
 
-            if (empty($live_endpoint_url_prefix)) {
+            $storedApikeyTest = Configuration::get('ADYEN_APIKEY_TEST');
+            // If input is empty AND database value is empty AND mode is test
+            if (empty($api_key_test) && empty($storedApikeyTest) && $mode === 'test') {
+                $output .= $this->displayError($this->l('Invalid Configuration value for test api key'));
```
acampos1916, line 889:
```suggestion
                $output .= $this->displayError($this->l('Invalid configuration value for test API key'));
```

adyenofficial.php
```diff
@@ -850,21 +872,43 @@ public function getContent()
...
+            }
+
+            $storedApikeyLive = Configuration::get('ADYEN_APIKEY_LIVE');
+            // If input is empty AND database value is empty AND mode is live
+            if (empty($api_key_live) && empty($storedApikeyLive) && $mode === 'live') {
+                $output .= $this->displayError($this->l('Invalid Configuration value for live api key'));
```
acampos1916, line 895:
```suggestion
                $output .= $this->displayError($this->l('Invalid configuration value for live API key'));
```

acampos1916 (dismissed):

views/js/adyen-admin.js
```diff
@@ -0,0 +1,64 @@
...
+            notificationSection.find('.control-label').removeClass('required');
+        }
+    }
+
+    /**
+     * Populate the api key inputs with some dummy asterisks for better visibility
+     */
+    function populateApiKeys() {
+        if (prodSection.find(":contains('Please fill your API key')").length === 0) {
+            prodSection.find(':password').val('****************');
```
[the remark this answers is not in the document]
Morerice: I tried this solution, but the password input field still ends up not being populated, which makes sense since normally one wouldn't want to pre-populate pw fields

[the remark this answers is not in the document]
cyattilakiss: in that case how do they know by default if the pw field is already filled?

adyenofficial.php
Morerice: https://github.com/Adyen/adyen-prestashop/pull/203/commits/14b3a2563093815d0d24e4c2a1a753c316b6e3e1

adyenofficial.php
Morerice: https://github.com/Adyen/adyen-prestashop/pull/203/commits/14b3a2563093815d0d24e4c2a1a753c316b6e3e1

cyattilakiss (approved):

acampos1916 (approved):

PR #207, opened 2021-06-14
[PW-4456] - Reduce the number of requests to /paymentMethods 

Morerice:
## Summary
The PrestaShop plugin makes unnecessary requests to the `/paymentMethods` Adyen endpoint. The requests are all done by the `fetchPaymentMethods` function in the adyenofficial file. This function is called by the following hooks:

* hookDisplayPaymentTop
* hookPaymentOptions (1.7)
* hookPayment (1.6)

Currently, for a single transaction the following api calls are done:

* 1.7 - **4 calls** to /paymentMethods
    * 1 in `hookDisplayPaymentTop`
    * 3 (1 for each of the 3 steps when already signed in) in `hookPaymentOptions`
* 1.6 (5-page checkout) - **2 calls** to /paymentMethods
    * 1 in `hookDisplayPaymentTop`
    * 1 in `hookPayment`
* 1.6 (1-page checkout) - **3 calls** to /paymentMethods
    * 2 in `hookDisplayPaymentTop`
    * 1 in `hookPayment`

After this PR is merged:

* 1.7 - **1 call** to /paymentMethods. 
    * 1 in `hookPaymentOptions`
* 1.6 (5-page checkout) - **1 call** to /paymentMethods.
    * 1 in `hookDisplayPaymentTop`
* 1.6 (1-page checkout) - **2 calls** to /paymentMethods
    * 2 in `hookDisplayPaymentTop` (since AdyenOfficial is instantiated on page load and on agreement to T&C).

## Tested scenarios
* All scenarios mentioned above
* E2E tests

### DIFF
--- a/adyenofficial.php
+++ b/adyenofficial.php
@@ -115,7 +115,7 @@ class AdyenOfficial extends PaymentModule
     private $checkout;
 
     /**
-     * @var \PrestaShop\PrestaShop\Core\Payment\PaymentOption[]
+     * @var array
      */
     private $paymentMethods;
 
@@ -1559,14 +1559,13 @@ class AdyenOfficial extends PaymentModule
             return array();
         }
 
-        //retrieve payment methods
-        $this->paymentMethods = $this->helper_data->fetchPaymentMethods($this->context->cart, $this->context->language);
+        $paymentMethods = $this->getPaymentMethods();
 
         if (!$this->context->customer->is_guest &&
-            !empty($this->paymentMethods['storedPaymentMethods']) &&
+            !empty($paymentMethods['storedPaymentMethods']) &&
             Configuration::get('ADYEN_ENABLE_STORED_PAYMENT_METHODS')
         ) {
-            $storedPaymentMethods = $this->paymentMethods['storedPaymentMethods'];
+            $storedPaymentMethods = $paymentMethods['storedPaymentMethods'];
             foreach ($storedPaymentMethods as $storedPaymentMethod) {
                 if (!empty($storedPaymentMethod)) {
                     // Only show on the frontend the Ecommerce stored payment methods and not the ContAuth
@@ -1608,8 +1607,8 @@ class AdyenOfficial extends PaymentModule
             }
         }
 
-        if (!empty($this->paymentMethods['paymentMethods'])) {
-            foreach ($this->paymentMethods['paymentMethods'] as $paymentMethod) {
+        if (!empty($paymentMethods['paymentMethods'])) {
+            foreach ($paymentMethods['paymentMethods'] as $paymentMethod) {
                 // Skip unsupported payment methods
                 if ($this->isUnsupportedPaymentMethod($paymentMethod['type'])) {
                     continue;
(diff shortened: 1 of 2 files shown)


acampos1916 (approved):

adyenofficial.php
```diff
@@ -2194,4 +2213,22 @@ private function getService($serviceName)
...
+
+    /**
+     * Check if payment methods have already been obtained in this object. If not, fetch them.
+     *
+     * @return array
+     * @throws Exception
+     */
+    private function getPaymentMethods()
+    {
+        if (empty($this->paymentMethods)) {
```
cyattilakiss, line 2225: In case the details changed like country code / amount / currency but the paymentMethods call has been done already, are we going to perform another payment methods again?

service/Checkout.php
```diff
@@ -24,10 +24,56 @@
 
 namespace Adyen\PrestaShop\service;
 
+use Adyen\PrestaShop\service\adapter\classes\ServiceLocator;
+use Cart;
+
 class Checkout extends \Adyen\Service\Checkout
 {
+    const PAYMENT_METHOD_STEP = 'checkout-payment-step';
```
cyattilakiss, line 32: do we have the same steps for one step checkout as well?

service/Checkout.php
```diff
@@ -24,10 +24,56 @@
...
+     * @return bool
+     */
+    public function isPaymentMethodStepNext(Cart $cart)
+    {
+        try {
+            $checkoutSessionData = \Db::getInstance()->getValue(
+                'SELECT checkout_session_data FROM ' . _DB_PREFIX_ . 'cart WHERE id_cart = ' . (int)$cart->id
+            );
+
+            $jsonData = json_decode($checkoutSessionData, true);
```
cyattilakiss, line 62: please add some defensive programming here

if the checkoutSessionData is empty then we can early return with the error
if the json_decode failed as well (try catch won't catch it) so you can do something similar to
```
if ($jsonData === null && json_last_error() !== JSON_ERROR_NONE) {
   echo "json data is incorrect";
}
```

service/Checkout.php
```diff
@@ -24,10 +24,56 @@
...
+
+            // If delivery step is complete and payment method is reachable
+            if ($jsonData[self::DELIVERY_STEP][self::IS_COMPLETE] &&
```
cyattilakiss, line 65: here as well. Always make sure that these keys `self::DELIVERY_STEP` `self::IS_COMPLETE` `self::PAYMENT_METHOD_STEP` `self::IS_REACHABLE` do exists otherwise it can create some nasty errors for you that will throw a warning and won't be caught by the try catch

service/Checkout.php
Morerice: On 1.7 the checkout is always considered as [one-page](https://www.prestashop.com/forums/topic/628709-enable-standard-5-step-checkout/) since technically you stay on the same controller while moving trough the steps. So the steps and the strings representing them will always be the same

adyenofficial.php
Morerice: Yes, in that case the payment methods will be fetched again, since the class is instantiated on every step. So if the user is in the payment methods page, and goes back to change something, class will be re-instantiated which would trigger the fetch call from Adyen

service/Checkout.php
Morerice: https://github.com/Adyen/adyen-prestashop/pull/207/commits/7a886af752f3521ac358b05509f8f9d2f2a9c41d

service/Checkout.php
Morerice: https://github.com/Adyen/adyen-prestashop/pull/207/commits/7a886af752f3521ac358b05509f8f9d2f2a9c41d

cyattilakiss (approved):

acampos1916 (approved):

PR #208, opened 2021-06-21
[PW-4877] - Create configuration for adyen styling

Morerice:
## Summary
A new configuration value to enable/disable the adyen styling should be created. When enabled, the adyen css file will be loaded whereas it won't be loaded if the config is disabled. This will help merchants that want to style the checkout page themselves, by not applying the styling of our files.

## Tested scenarios
* Config enabled
* Config disabled

### DIFF
--- a/adyenofficial.php
+++ b/adyenofficial.php
@@ -871,6 +871,7 @@ class AdyenOfficial extends PaymentModule
             $google_pay_merchant_identifier = Tools::getValue('ADYEN_GOOGLE_PAY_MERCHANT_IDENTIFIER');
             $payment_display_collapse = Tools::getValue('ADYEN_PAYMENT_DISPLAY_COLLAPSE');
             $enable_stored_payment_methods = Tools::getValue('ADYEN_ENABLE_STORED_PAYMENT_METHODS');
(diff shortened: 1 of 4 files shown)


acampos1916 (approved):

adyenofficial.php
```diff
@@ -1646,7 +1684,11 @@ public function hookPayment()
             return null;
         }
 
-        $this->context->controller->addCSS('modules/' . $this->name . '/views/css/adyen.css', 'all');
+        $loadAdyenCss = Configuration::get('ADYEN_ENABLE_CHECKOUT_STYLING');
+
+        if ($loadAdyenCss) {
+            $this->context->controller->addCSS('modules/' . $this->name . '/views/css/adyen.css', 'all');
```
cyattilakiss, line 1690: is this going to disable the checkout component css as well or only the ones that we added as extra for prestashop adjustments?

cyattilakiss (approved):

adyenofficial.php
Morerice: Nope. The file which will not be imported only contains css specific for the prestashop plugin

Issue #192, opened 2021-05-05
adyen change status order prestashop before payment is finisht


ikkiedikkie:
Hello Support,

There is a problem in using adyen with prestashop 1.7.7.2

See the order from 1060€ it is open so the order must not change the staus in prestashop
![1](https://user-images.githubusercontent.com/83693470/117108513-9a8a2d80-ad83-11eb-8b61-784bfba934af.png)

But anyway the order is changed
![2](https://user-images.githubusercontent.com/83693470/117108515-9b22c400-ad83-11eb-8a06-350edc6717a4.png)

It may never changed only if there is a payment

But ok the biggest problem is that in prestashop the order is declared as payed!!!!!! And that is False.
![3](https://user-images.githubusercontent.com/83693470/117108516-9b22c400-ad83-11eb-8bc0-ef47af3f86b0.png)
 

This is a wrong workaround from Adyen.

prestashop 1.7.7.2
chome Versie 90.0.4430.93

Morerice:
Hi @ikkiedikkie 

Since the issue is linked to an already live merchant and there aren't any technical details provided, could you please contact our [support team](https://www.adyen.help/hc/en-us/requests/new?ticket_form_id=360000705420)? 

Best regards,
Jean
Adyen

ikkiedikkie:
> Hi @ikkiedikkie
> 
> Since the issue is linked to an already live merchant and there aren't any technical details provided, could you please contact our [support team](https://www.adyen.help/hc/en-us/requests/new?ticket_form_id=360000705420)?
> 
> Best regards,
> Jean
> Adyen

Hi Jean,

i did and the support team send me back to github.
what more technical details do you need so i can give them.

best regards ikkie

Morerice:
> > Hi @ikkiedikkie
> > Since the issue is linked to an already live merchant and there aren't any technical details provided, could you please contact our [support team](https://www.adyen.help/hc/en-us/requests/new?ticket_form_id=360000705420)?
> > Best regards,
> > Jean
> > Adyen
> 
> Hi Jean,
> 
> i did and the support team send me back to github.
> what more technical details do you need so i can give them.
> 
> best regards ikkie

Hi @ikkiedikkie,

Which plugin version are you currently on? Also, could you please provide us with the [adyen](https://docs.adyen.com/plugins/prestashop#finding-the-logs) log files?

Best regards,
Jean
Adyen

ikkiedikkie:
[...]

PR #214, opened 2021-07-08
[PW-4943] - Update logable value of our WAITING_FOR_PAYMENT state

Morerice:
## Summary
The **WAITING_FOR_PAYMENT** order state should have a logable value of 0. This should be done since when the order is in this status, no payment has been processed yet and hence, no OrderPayment linked to this order should be created. This will also ensure that the `total_paid_real` db value, will only be updated once the payment actually goes trough, and not during the redirection.

**Steps:**
1. When creating the WAITING_FOR_PAYMENT state, set the `logable` value to 0.
3. Create upgrade script

## Tested scenarios
* `total_paid_real = 0` during redirection on 1.6
* `total_paid_real = 0` during redirection on 1.7

## Fixed issues

#192

### DIFF
--- a/upgrade/upgrade-3.7.1.php
+++ b/upgrade/upgrade-3.7.1.php
@@ -68,7 +68,7 @@ function set_waiting_for_payment_status_logable_to_false()
 
     return Db::getInstance(_PS_USE_SQL_SLAVE_)->update(
         'order_state',
-        ['logable' => false],
+        array('logable' => false),
         'id_order_state = ' . $orderStateConfigurationId,
         1
     );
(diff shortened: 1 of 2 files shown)


upgrade/upgrade-3.7.1.php
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    if (empty($orderState)) {
+        return false;
+    }
+
+    return Db::getInstance(_PS_USE_SQL_SLAVE_)->update(
+        'order_state',
+        array('logable' => false),
+        'id_order_state = ' . $orderStateConfigurationId,
```
cyattilakiss, line 72:
```suggestion
        'id_order_state = ' . (int)$orderStateConfigurationId,
```

acampos1916 (dismissed):

cyattilakiss (approved):

acampos1916 (approved):

PR #235, opened 2022-01-05
Add hook to NotificationProcessor::processNotification

Shagshag:
## Summary
Try to process the notification in other modules before doing the default process. Usefull for partial refunds or additional payments that the module does not manage

### DIFF
--- a/service/notification/NotificationProcessor.php
+++ b/service/notification/NotificationProcessor.php
@@ -150,6 +150,18 @@ class NotificationProcessor
      */
     public function processNotification($unprocessedNotification)
     {
+        // try to process the notification in other modules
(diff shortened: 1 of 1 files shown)


pkrstojevic-logeecom (approved):

logeecom:
Hi @Shagshag,
 
Thank you very much for providing us with a suggested hook for notifications processing. 

However, the suggested change from PR will introduce dependencies towards third parties for notification processing, which might have a negative impact on module support in cases when notifications are not processed as expected. 

Considering previously, our development team will have to ensure that third parties are not overtaking the notification processing and consider this change put in the backlog as a feature request.

Kind regards.

PR #237, opened 2022-01-12
[PW-6058] - Fetch payment methods on virtual cart

Morerice:
## Summary
Currently when a shopper is checking out with a virtual cart (all products to be bought are virtual) the adyen payment methods are not being fetched and displayed. This is due to a check on the delivery step of the checkout process, which on a virtual cart will not exist.

This PR will solve this issue by checking if the cart is virtual and if so, omitting the check on the delivery step.

## Tested scenarios
* Checkout with a virtual cart
* Checkout with a normal cart
* Checkout with a mixed cart (1 virtual, 1 normal)

### DIFF
--- a/service/Checkout.php
+++ b/service/Checkout.php
@@ -56,6 +56,7 @@ class Checkout extends \Adyen\Service\Checkout
      */
     public function requireFetchPaymentMethods(Cart $cart)
     {
+        $isVirtualCart = $cart->isVirtualCart();
         try {
             $checkoutSessionData = \Db::getInstance()->getValue(
                 'SELECT checkout_session_data FROM ' . _DB_PREFIX_ . 'cart WHERE id_cart = ' . (int)$cart->id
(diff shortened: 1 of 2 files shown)


service/Checkout.php
```diff
@@ -76,11 +77,11 @@ public function requireFetchPaymentMethods(Cart $cart)
             }
 
             // Check if all keys exist, and then check if the delivery step is complete and payment method is reachable
-            if (array_key_exists(self::DELIVERY_STEP, $jsonData) &&
+            if (($isVirtualCart || array_key_exists(self::DELIVERY_STEP, $jsonData)) &&
                 array_key_exists(self::IS_COMPLETE, $jsonData[self::DELIVERY_STEP]) &&
```
peterojo, line 81: line 81 would break (undefined index) if `$isVirtualCart` is `true` and `self::DELIVERY_STEP` does not exist in the `$jsonData` array

peterojo: there should be a simpler way to do this using `isset()` (like `isset($jsonData[self::DELIVERY_STEP][self::IS_COMPLETE])`) or better still using null coalesce operator (if you only support PHP 7 and above)

Morerice: Updated and cleaned up the if condition a bit in https://github.com/Adyen/adyen-prestashop/pull/237/commits/ded6de69944df0d6948413cc161ced4685ed338c

peterojo (approved):

PR #261, opened 2023-06-20
prestashop 9 logger addRecord bug fix for Monolog compatibility

Jovonni:
## Summary
Describe the changes proposed in this pull request:
- What is the motivation for this change? 
Prestashop 9 throws an error from Logger->addRecord
```php
Compile Error: Declaration of Adyen\PrestaShop\service\Logger::addRecord($level, $message, array $context = [], $callPrestaShopLogger = true) 
must be compatible with Monolog\Logger::addRecord(int $level, string $message, array $context = [], ?Monolog\DateTimeImmutable $datetime = null): bool
```
- What existing problem does this pull request solve?
Prestashop 9 compatibility

## Tested scenarios

**Fixed issue**:  
Fix:
```php
public function addRecord(int $level, string $message, array $context = [], ?\Monolog\DateTimeImmutable $datetime = null): bool
    {
        $context['is_exception'] = $message instanceof \Exception;
        if (array_key_exists($level, self::$prestashopLoggable)) {
            \PrestaShopLogger::addLog(
                $message,
                self::$prestashopLoggable[$level],
                null,
                null,
                null,
                true
            );
        }

        return parent::addRecord($level, $message, $context, $datetime);
    }
```

### DIFF
--- a/service/Logger.php
+++ b/service/Logger.php
@@ -197,10 +197,10 @@ class Logger extends \Monolog\Logger
      *
      * @return bool Whether the record has been processed
      */
-    public function addRecord($level, $message, array $context = [], $callPrestaShopLogger = true)
+    public function addRecord(int $level, string $message, array $context = [], ?\Monolog\DateTimeImmutable $datetime = null): bool
     {
         $context['is_exception'] = $message instanceof \Exception;
-        if (array_key_exists($level, self::$prestashopLoggable) && $callPrestaShopLogger) {
+        if (array_key_exists($level, self::$prestashopLoggable)) {
             \PrestaShopLogger::addLog(
                 $message,
                 self::$prestashopLoggable[$level],
@@ -211,7 +211,7 @@ class Logger extends \Monolog\Logger
             );
         }
 
-        return parent::addRecord($level, $message, $context);
+        return parent::addRecord($level, $message, $context, $datetime);
     }
 
     /**


Jovonni:
fyi haven't tested thoroughly, but it solves the issue in prestashop.

Jovonni:
quite a few errors with prestashop 9

DamjanLogeecom:
Hello @Jovonni,

Thank you for the feedback,

Our team will review your proposed changes and we will update you once we have more information.

Kind regards

DamjanLogeecom:
Hello @Jovonni,

Considering the unavailability of PrestaShop version 9 at the moment, we regret to inform you that we are unable to replicate or test the reported issue.

Upon thorough examination of the pull request and an investigation into the PrestaShop logger (in previous versions), we have determined that the proposed changes are not compatible with the supported PrestaShop versions for the Adyen plugin (versions 1.7.0.0 - 8.0.4).

In these versions, the Monolog/Logger method, specifically the addRecord method, consists of three parameters: logging level (integer), message (string), and log context (array). However, the pull request introduces a fourth parameter, ?\Monolog\DateTimeImmutable $datetime. Additionally, the pull request incorporates type declarations in the method arguments and specifies the return type of the method, which is only available starting from PHP 7.4 (as outlined in the PHP documentation on type declarations: https://www.php.net/manual/en/language.types.declarations.php ). Given that the majority of PrestaShop versions compatible with the Adyen plugin support earlier versions of PHP, this change cannot be implemented.

Further investigation into this issue should be conducted once PrestaShop version 9 becomes available.

Kind regards.

PR #284, opened 2024-12-10
Fix - Cannot declare class Autoloader, because the name is already in use

sergio-sanchis:
## Summary
When trying to update the module from a version prior to 5.2.0 I got an error that the Autoloader class was already in use because another module was using the same class name.

## Tested scenarios
Having an Adyen version prior to 5.2.0 with the Doofinder module, which is the one we have detected the conflict with.

When trying to update Adyen to the latest version, the error occurred.
"Cannot declare class Autoloader, because the name is already in use"

### DIFF
--- a/src/upgrade/Autoloader.php
+++ b/src/upgrade/Autoloader.php
@@ -1,5 +1,7 @@
 <?php
 
+namespace AdyenPayment;
+
 class Autoloader
 {
     /**
--- a/src/upgrade/upgrade-5.0.0.php
+++ b/src/upgrade/upgrade-5.0.0.php
@@ -46,8 +46,8 @@ require_once 'Autoloader.php';
  */
 function upgrade_module_5_0_0(AdyenOfficial $module): bool
 {
-    Autoloader::setFileExt('.php');
-    spl_autoload_register('Autoloader::loader');
+    \AdyenPayment\Autoloader::setFileExt('.php');
+    spl_autoload_register('\AdyenPayment\Autoloader::loader');
 
     $previousShopContext = Shop::getContext();
     Shop::setContext(ShopCore::CONTEXT_ALL);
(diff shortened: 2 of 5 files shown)


adyen-integrations-support:
Hey @sergio-sanchis ,

Thank you for reaching out!

Our team will check this and get back to you.

Best regards.

adyen-integrations-support:
Hello @sergio-sanchis ,

We attempted to reproduce the issue but were unable to do so. We installed PrestaShop version 1.7.8.11 along with the plugin version 5.2.0 and successfully updated it to the latest version without encountering any problems. Additionally, the Doofinder module was installed and connected during the process, and this did not affect the plugin update in any way.
Screenshot from 2025-01-10 14-28-40

We also tried with PrestaShop 8.2 and Plugin Version 5.2.4: During testing, the system required selecting an update mode: either "Maintenance" (via the Maintenance page) or "Normal" (Upgrade anyway). Both modes were tested, and the plugin was updated successfully in each case.
Screenshot from 2025-01-10 13-56-13
We used this plugin for testing. Could you please confirm if this is the same plugin you are using?https://addons.prestashop.com/en/search-filters/30818-doofinder-search-discovery.html 

Could you please provide us with additional information listed below?

- Description* - Describes in detail what is the problem or error.
- Steps to reproduce - Information on how to reproduce a reported issue with detailed steps.
- Screenshots - Screenshots that describe the issue.
- System version* - System version.
- Plugin version* - Plugin version installed to the integrated system.

After that, we will check all reported issues immediately and come back to you with feedback as soon as possible.

Kind Regards.

adyen-integrations-support:
Hi @sergio-sanchis ,

We wanted to check in and see if this issue has been resolved. Are you still experiencing any problems?

If the issue is resolved, could you please confirm so we can proceed with closing the ticket?

Thank you in advance.

Best regards

